PR #415, opened 2014-10-01
Set phantomjs viewportSize to 1024x768

sebastianseilund:
As discussed in https://github.com/airportyh/testem/pull/398 this is a more common screen size that the default 400x300.

### DIFF
--- a/assets/phantom.js
+++ b/assets/phantom.js
@@ -1,3 +1,7 @@
 var page = require('webpage').create();
 var url = phantom.args[0];
+page.viewportSize = {
+  width: 1024,
+  height: 768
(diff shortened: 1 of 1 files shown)


williamsbdev:
This seems reasonable, is there any way for us to be able to override this though? I just upgraded to Phantomjs 2.1.1 from 1.9.8 and because the viewportSize changed, we have failing tests (as we were incorrectly expecting the smaller window for our media queries). Is there going to be some way to set the viewportSize per test or per run?

johanneswuerbach:
Currently there isn't, please open a new issue for that :-)

PR #441, opened 2014-12-09
added new option to change TEMP dir; added properties that points to user TEMP dir in win32

my8bit:


my8bit:
Hi Toby! Can you please comment what is wrong with this pull request? I guess, at least, [adding](../commit/61ab0e4c6b8e765a3460db63e333ab3db403ef99) more accurate detection of temporary environment variables to will be a good feature.

airportyh:
`fs.lstatSync` call is a synchronous IO function. I don't want any of those in the code base.

Issue #452, opened 2015-01-07
Broken IE 8 / Slow script warning


johanneswuerbach:
It looks like https://github.com/airportyh/testem/pull/450 broke running tests in IE 8. 

https://saucelabs.com/tests/509a6e5accb64ed7bffb35cd1b2c1e1b
"Stop running this script?" error dialog in the end of the screencast.
https://www.dropbox.com/s/fn0m9dy7loonsr7/Screenshot%202015-01-07%2011.24.07.png?dl=0

I'll investigate asap, but I need to get my hands on an IE 8 VM.

johanneswuerbach:
It looks like the [decycling](https://github.com/airportyh/testem/blob/master/public/testem/testem_client.js#L272) is really slow now.

https://www.dropbox.com/s/xqfifka0s55xsda/Screenshot%202015-01-07%2015.24.34.png?dl=0

johanneswuerbach:
It looks like the error reporting was broken in IE 8 before and the error just changed by https://github.com/airportyh/testem/commit/9991e29d9601f4de23046c37b70e65f2b2ef39d2

PR #453, opened 2015-01-08
Fixed test running in IE 8

johanneswuerbach:
Faster de-cycling (no more stringify & parse, max depth)
Unit: Only send expected / actual when test failed

Fixes https://github.com/airportyh/testem/issues/452


airportyh:
Thanks!

PR #455, opened 2015-01-14
Prevent XMLHttpRequest in socket.io from being faked

Dremora:
[Pretender](https://github.com/trek/pretender) (amongst others) swaps out window.XMLHttpRequest for a fake XHR object and undoes the fake at the end of each test. This obviously presents an issue for async tests where Testem might want to communicate over XHR midway through a test. It will encounter a fake XHR and things will stop being fun.

This patch keeps a reference to the original `window.XMLHttpRequest` as a variable within the lexical scope of the module that uses it.

See this [test case](https://github.com/jgwhite/testem-case) with ember-cli + pretender for a reproduction.

Run `ember test --server` in the repo to see Firefox get stuck in an reload loop. This patch should fix that.

### DIFF
--- a/public/testem/socket.io.js
+++ b/public/testem/socket.io.js
@@ -3370,6 +3370,7 @@ WS.prototype.check = function(){
 },{"../transport":13,"component-inherit":20,"debug":21,"engine.io-parser":24,"parseqs":32,"ws":34}],19:[function(_dereq_,module,exports){
 // browser shim for xmlhttprequest module
 var hasCORS = _dereq_('has-cors');
+var XMLHttpRequest = window.XMLHttpRequest;
 
 module.exports = function(opts) {
   var xdomain = opts.xdomain;


johanneswuerbach:
I think it's better when you patch this outside `socket.io.js` as I would like to remove this file as soon as a new `socket.io` version is released.

Just put a new file into `public/testem` and add it here: https://github.com/airportyh/testem/blob/master/lib/server/index.js#L196

Dremora:
I can't think of a way to apply this patch from outside of `socket.io.js`. One of its modules exports a function which refers to `XMLHttpRequest`. All I did was to create a local binding to the global variable within the module, but outside the function. As a result of this, local `XMLHttpRequest` variable is evaluated only once and always refers to whatever `window.XMLHttpRequest` was referring to at the time of module evaluation.

However, I agree that maintaining a patched version of socket.io is not the clearest thing in the world. There are two solutions we can go for to mitigate the issue:

1. Submit a patch for socket.io itself. This, however, will prevent `XMLHttpRequest` from being faked even when needed - i.e., in cases when the project we are testing relies on socket.io and we do want to fake it.
2. Use `pretender.passthrough` to always use native XMLHttpRequest for a subset of URLs used to communicate with testem. This is probably the cleanest solution.

jgwhite:
It seems like we should pursue both solutions.

For the next move, I suggest a documentation PR to [trek/pretender](https://github.com/trek/pretender) explaining how to mitigate this problem using a passthrough for `/socket.io`.

johanneswuerbach:
@Dremora I PR-ed a solution, which doesn't require patching and tested it with your repo. https://github.com/airportyh/testem/pull/457 Does this work for you?

jgwhite:
@johanneswuerbach looks good :+1:

jgwhite:
@Dremora in light of #457 you can close this PR. Glad we were able to solve this cleanly.

PR #461, opened 2015-01-20
QUnit 1.16.0

mkoxio:
Some issues has been reported(jquery/qunit#732 and jquery/qunit#733) of new version 1.17 cause some breaks under phantomJS, so I think it's safe to use 1.16.0 for now

### DIFF
--- a/examples/qunit_simple/tests.js
+++ b/examples/qunit_simple/tests.js
@@ -1,8 +1,8 @@
-test('says hello world', function(){
+QUnit.test('says hello world', function(){
     equal(hello(), 'hello world', 'should equal hello world');
 });
 
-test('says hello to person', function(){
+QUnit.test('says hello to person', function(){
     equal(hello('Bob'), 'hello Bob', 'should equal hello Bob');
 });
 
--- a/public/testem/qunit_adapter.js
+++ b/public/testem/qunit_adapter.js
@@ -5,8 +5,8 @@ qunit_adapter.js
 
 Testem's QUnit adapter. Works by using QUnit's hooks:
 
-* `testStart`
-* `testDone`
+* `begin`
+* `done`
 * `moduleStart`
 * `moduleEnd`
 * `done`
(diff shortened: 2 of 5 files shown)


public/testem/qunit_adapter.js
```diff
@@ -5,8 +5,8 @@ qunit_adapter.js
 
 Testem's QUnit adapter. Works by using QUnit's hooks:
 
-* `testStart`
-* `testDone`
+* `begin`
+* `done`
```
leobalter: these changes might break.

`testStart` and `testDone` are hooks for before and after each test block, while `begin` and `done` stands for the entire suite.

views/qunitrunner.mustache
```diff
@@ -9,11 +9,7 @@
 {{#styles}}<link rel="stylesheet" href="{{.}}">{{/styles}}
 </head>
 <body>
-<h1 id="qunit-header">Tests</h1>
-<h2 id="qunit-banner"></h2>
-<div id="qunit-testrunner-toolbar"></div>
-<h2 id="qunit-userAgent"></h2>
-<ol id="qunit-tests"></ol>
+<div id="qunit"></div>
```
leobalter: as long as you use `<div id="qunit"></div>` as proposed, QUnit 1.17.0 won't break anything on PhantomJS. Anyway, it's worth waiting 1 day to get 1.17.1 fixing this minor regression.

examples/qunit_simple/tests.js
```diff
@@ -1,8 +1,8 @@
-test('says hello world', function(){
+QUnit.test('says hello world', function(){
```
leobalter: this change is not necessary yet as the global methods like `test`, `module`, etc will still be supported until QUnit 2.0.0.

Even though, if you like them namespaced, as I, I would suggest to use the assert argument, example:

```js
QUnit.test('says hello world', function(assert){
    assert.equal(hello(), 'hello world', 'should equal hello world');
```

public/testem/qunit_adapter.js
airportyh: That might be a problem. IIRC Testem needs to hook into each test to report test results incrementally.

views/qunitrunner.mustache
leobalter: QUnit 1.17.1 is now released.

PR #460, opened 2015-01-20
QUnit 1.17.0

mkoxio:
#456  some tests fails even before I made the changes locally on my mac, but I made this PR anyway, hopefully doesn't break anything in the travis-CI

PR #470, opened 2015-01-22
Qunit 1.17.1

mkoxio:
Update to 1.17.1, this should fix the bug under phantomJS. and also change the related file as you guys commented on #461 and #460


johanneswuerbach:
Thank you.

PR #471, opened 2015-01-26
Got rid of comma-first style - I don't even do this anymore and I can im...

airportyh:
...agine it is off putting to contributors. Also some minor clean up.


piuccio:
nice

PR #472, opened 2015-01-27
added pause functionality in TDD mode as well as specs for this functionality

kbullaughey:
Now in TDD mode the person can type `p` and it will pause testem so that it doesn't run any tests (either due to file changes or when resetting testem). Pressing `p` again will unpause testem. This is useful if one is going to be changing lots of files and doesn't want to completely halt testem but doesn't want it running all the specs.


airportyh:
Cool! Thanks.

PR #473, opened 2015-01-31
Get the temp dir path from env

my8bit:
Get the temporary directory path from enviroment variables if exist.


johanneswuerbach:
Thanks for the PR :+1: . Could you change this to http://nodejs.org/api/os.html#os_os_tmpdir? It uses your proposed logic, so we don't need to maintain a separate implementation. https://github.com/joyent/node/blob/v0.10.36/lib/os.js#L44

johanneswuerbach:
Thanks for the change, could you squash your commits into a single one?

johanneswuerbach:
Looks like there are 4 commits now: https://github.com/airportyh/testem/pull/473/commits, but it should be only one. I can do it for you, if you like.

PR #477, opened 2015-02-03
Get the temp dir path from env from os.tmpdir()

johanneswuerbach:
Replaces https://github.com/airportyh/testem/pull/473


my8bit:
Many thanks!

johanneswuerbach:
Thank you. Your change is released in 0.6.36

PR #482, opened 2015-02-05
Removed unused files

johanneswuerbach:
I couldn't find any place where those files are used.

@airportyh r?


airportyh:
It's good :)

PR #486, opened 2015-02-07
find phantomjs by node module resolution

jsturgis:
Add the ability to find phantomjs by node module resolution if which fails. This will allow people to have phantomjs installed in their local node_modules folder instead of requiring it to be installed globally.

### DIFF
--- a/lib/fileutils.js
+++ b/lib/fileutils.js
@@ -18,6 +18,9 @@ exports.findableByWhere = findableByWhere
 var findableByWhich = findableBy(which)
 exports.findableByWhich = findableByWhich
 
+exports.findableByWhichOrModule = findableByTypeOrModule(which)
+exports.findableByWhereOrModule = findableByTypeOrModule(where)
+
 function findableBy(func, cb){
   return function(cb){
     var browser = this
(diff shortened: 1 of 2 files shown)


lib/fileutils.js
```diff
@@ -43,3 +68,29 @@ function which(exe, cb){
     cb(!!exePath)
   })
 }
+
+exports.findableByWhichOrModule = findableByWhichOrModule
+function findableByWhichOrModule(cb){
+  var browser = this
+
+  findableByWhich.call(browser, function(findable){
```
johanneswuerbach: I think the local version should be preferred over the global version. So overwriting works on a per project basis.

lib/fileutils.js
```diff
@@ -43,3 +68,29 @@ function which(exe, cb){
...
+
+exports.findableByWhichOrModule = findableByWhichOrModule
+function findableByWhichOrModule(cb){
+  var browser = this
+
+  findableBy(which).call(browser, function(findable){
+    if(findable){
+      cb(findable)
+    }else{
+      findableByModule.call(browser, cb)
```
johanneswuerbach: Shouldn't we look first for the module and then for the exe to find a local phantomjs, before finding the global one?

lib/fileutils.js
```diff
@@ -43,3 +68,29 @@ function which(exe, cb){
...
+    }
+  })
+}
+
+exports.findableByWhereOrModule = findableByWhereOrModule
+function findableByWhereOrModule(cb){
```
johanneswuerbach: Could you merge the common parts with `findableByWhichOrModule`? Both look nearly the same and it is difficult to spot the difference.

johanneswuerbach:
Thanks for the update :+1: .Please ping me when you updated the PR. GitHub doesn't send any notifications for code changes :cry:

jsturgis:
Ok no problem I'll update you when I fix the issues you mentioned.

jsturgis:
@johanneswuerbach I made those changes you requested can you take a look and let me know anything else I should change?

johanneswuerbach:
This looks awesome. As a last step, could you squash your commits into a single one? http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit

jsturgisRN:
@johanneswuerbach I squashed those commits, I am seeing this error in the AppVeyor test:
"npm ERR! registry error parsing json"

Any idea what might cause that?

johanneswuerbach:
Thank you @jsturgisRN :+1: AppVeyor was never passing, but I'm working on it.

PR #487, opened 2015-02-07
Proposal: Separate testem connection into an iframe

johanneswuerbach:
Prevent leaking code from testem into tested application and vice versa

This prevents issues like https://github.com/airportyh/testem/pull/455 in the furture.

@airportyh I tried to introduce no breaking changes, but I think this should be released as 0.7. What do you think?


airportyh:
Yeah, bumping the minor version is good as it's somewhat risky and may take a while to stabilize.

Issue #437, opened 2014-12-02
subshells missing user options


jasonkarns:
Is it intentional that commands (as specified in testem.json) run in subshells with differing environments?

Options set via a user's .profile or .bashrc, for instance, are not in effect when testem's commands are run (like `before_tests` for example).

jasonkarns:
I opened this issue specifically for shell options not being the same in the testem-spawned child processes. According to the bash reference, https://www.gnu.org/software/bash/manual/html_node/Command-Execution-Environment.html, shell options are *not* copied over the child processes. So I'm going to consider this closed as far as testem is concerned.

PR #493, opened 2015-02-09
Ensure that environment is copied to sub commands

johanneswuerbach:
Fixes https://github.com/airportyh/testem/issues/437


jasonkarns:
this doesn't actually fix #437

I tested by making custom launcher:

``` json
{
  "launchers": {
    "shopt": {
      "command": "shopt"
    }
  },
  "launch_in_dev": ["shopt"]
}
```

The output of `shopt` shows which shell options are enabled/disabled, and they are certainly not the same as the shell that launched testem. However, I'm not convinced they *ought* to be, so I'm closing #437 (more details there)

PR #505, opened 2015-02-20
WIP: Windows and PhantomJS 2.0

johanneswuerbach:
Chocolatey is installing PhantomJS now inside %PATH%

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ install:
   - npm install
   # Install PhantomJS
   - IF %PHANTOM_SRC%==npm npm install -g phantomjs
-  - IF %PHANTOM_SRC%==chocolatey cinst PhantomJS -Version 1.9.8 -y && set path=%path%;C:\tools\PhantomJS\
+  - IF %PHANTOM_SRC%==chocolatey cinst PhantomJS -y
   - phantomjs -v
 
 # Post-install test scripts.


johanneswuerbach:
The issue is that chocolatey is using a shim now to exec phantomjs. After testem received all test results from PhantomJS, it tries to kill the process, but only has the PID of the shim, which already exited.

I don't have an idea how to workaround this atm. :disappointed: 

//cc @stefanpenner

stefanpenner:
ahh, can we pull in someone from the chocolatey team?

johanneswuerbach:
Maybe @ferventcoder or @Iristyle? Is there a way to keep the shim running and kill the tree?

ferventcoder:
Help me understand what you want to do. It sounds like this tool should kill phantomjs.exe (the actual, not the shim). The shim should be blocking until phantomjs shuts down.

ferventcoder:
I'm guessing you kill the shim and you want to kill the entire tree. I think you already stated that.

As you can see here the shim is blocking. It won't exit on it's own until the phantomjs.exe (actual) exits.
image

johanneswuerbach:
Currently I'm sending SIGTERM to the process with the pid I spawned (which I guess is the shim). This seems to kill the shim, but not phantomjs. When I'm installing phantomjs manually or use 1.9.8 this works as expected.

ferventcoder:
I wonder if we can enhance shims that receive a SIGTERM to destroy the tree? That seems like a much better option.

ferventcoder:
I created https://github.com/chocolatey/shimgen/issues/11 to handle this.

johanneswuerbach:
That sounds great. Here is a small example script to reproduce the issue:
```js
var cp = require('child_process');
var p = cp.spawn('phantomjs', ['--webdriver=8910'], {}); // The wd arg just keeps phantom listening
p.on('close', function() {
  console.log('Should be called.')
});
setTimeout(function() {
  p.kill();
}, 2000);
```

ferventcoder:
Thanks. I will have something out after a bit.

ferventcoder:
So I can't find a nice way to handle SIGKILL but SIGTERM is going to be in Chocolatey 0.9.9 (you will need to regenerate your shims). It will be out in rc8 (due out this weekend).

ferventcoder:
I take it you are good to go? :)

johanneswuerbach:
Yes, I had to switch from `process.kill` to executing `taskkill /T` on Windows as of https://github.com/airportyh/testem/pull/520 (npm is creating a cmd file when installing phantomjs), which also solves this issue.

PR #506, opened 2015-02-20
Opera and socket fix

johanneswuerbach:


johanneswuerbach:
@MoOx this does not fix the race, but fixes opera for me.

MoOx:
:+1:

Issue #529, opened 2015-04-01
XUnit reporter fails if fields are null


Twinkletoes:
PR on its way with new test added

PR #531, opened 2015-04-01
XUnit CI reporter uses xmldom

Twinkletoes:
Fixes https://github.com/airportyh/testem/issues/529

Use XML DOM rather than string concatenation. Also make xml schema correct (missing required attributes). Augmented readme with example output


stefanpenner:
very nice!

PR #533, opened 2015-04-03
Handle net.Server.close() throws "Not running"

yayalice:
For backwards compatability with node v0.11.13 and earlier

This can occur when testem tries to connect with a port that is already in use. This was exposed by https://github.com/airportyh/testem/commit/0c63c8

Alternatively, we can modify the README to state node >= 0.11.14 is required (currently it's at 0.6.2)

### DIFF
--- a/lib/ci/index.js
+++ b/lib/ci/index.js
@@ -147,6 +147,10 @@ App.prototype = {
     this.reporter.finish()
     this.emit('tests-finish')
     this.stopHookRunners()
+    if (err && /^Testem Server Error/.test(err.message)){
+      this.stopped = true
+      return this.exit()
+    }
     this.stopServer(this.exit.bind(this))
   },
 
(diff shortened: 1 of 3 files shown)


johanneswuerbach:
Testem was never actually tested against unstable versions of node and I think supporting 0.11 now that 0.12 was released is not worth the effort.

Is there a specific reason you can't use 0.12?

johanneswuerbach:
Actually, this function shouldn't be called when the server is not running. Lets fix this, which will also bring back 0.11 compatibility.

yayalice:
Updated PR - let me know if this is what you meant. I actually ran into this bug with node 0.10.

johanneswuerbach:
Thanks, looks good now. How can I create an error like this?

yayalice:
I got it by setting testem's config to use a port that's already in use, with node 0.10.28


On Saturday, Apr 4, 2015 at 01:42, Johannes Würbach <notifications@github.com>, wrote:

Thanks, looks good now. How can I create an error like this?

PR #535, opened 2015-04-07
Fix multiple content-type proxying.

ehntoo:
I tried using the example for proxying multiple content types in README.md and it wasn't working. Looks like there was a small mix-up in the implementation.

### DIFF
--- a/tests/server_tests.js
+++ b/tests/server_tests.js
@@ -33,7 +33,7 @@ describe('Server', function(){
           },
           '/api3': {
             target: 'http://localhost:13374',
-            onlyContentTypes: ['json']
+            onlyContentTypes: ['json', 'xml']
           },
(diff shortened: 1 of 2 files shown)


tests/server_tests.js
```diff
@@ -33,7 +33,7 @@ describe('Server', function(){
           },
           '/api3': {
             target: 'http://localhost:13374',
-            onlyContentTypes: ['json']
+            onlyContentTypes: ['json', 'xml']
```
johanneswuerbach: Shouldn't `api3` not handle `xml`? I think the test title is misleading.

johanneswuerbach:
Closed as no feedback recei

johanneswuerbach:
Friendly ping :-)

PR #538, opened 2015-04-10
Height/weight styling for connection.html

mkeemon:
When connection.html is loaded in an iframe, it inherits the height/width from the parent element, causing it to take up the entire page. This prevents users from clicking anything on the page. This just adds very basic styling to prevent that from happening.

### DIFF
--- a/public/testem/connection.html
+++ b/public/testem/connection.html
@@ -12,6 +12,8 @@
     position: fixed;
     bottom: 5px;
     right: 5px;
+    height: 85px;
+    width: 180px;
     background-color: #444;
     padding: 3px;
     color: #fff;


package.json
```diff
@@ -7,7 +7,7 @@
     "unittest",
     "browser"
   ],
-  "version": "0.8.0-0",
+  "version": "0.8.0-1",
```
johanneswuerbach: Please do not bump the version.

johanneswuerbach:
Do you have an example main page css, which can replicate the issue?

mkeemon:
After further investigation, there was css in the application we were testing that was modifying iframe styling. Sorry about that!

PR #543, opened 2015-04-20
Add teamcity reporter

jbryson3:
The tap reporter has become a bit of a pain for us, now that we've upgraded to Teamcity 9.  I added a teamcity reporter and tested it with our stack. It's pretty basic, but should be a good starting point for us and others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ TAP is a human-readable and language-agnostic test result format. TAP plugins ex
 
 ## Other Test Reporters
 
-Testem has other test reporters than TAP: `dot` and `xunit`. You can use the `-R` to specify them
+Testem has other test reporters than TAP: `dot`, `xunit` and `teamcity`. You can use the `-R` to specify them
 
     testem ci -R dot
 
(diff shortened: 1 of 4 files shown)


johanneswuerbach:
Restarted windows ci.

johanneswuerbach:
Could you add the reporter the README?

jbryson3:
Good catch, I knew I forgot to do something.  I added sample output to the README as well. Not sure if that's helpful since it should just plugin to Teamcity's api.  I bumped the version as well, I never know whether to do that for a pull request or not.  I tried to break up the commits so you could selectively pull what you wanted. Thanks!

johanneswuerbach:
Thank you, version bumping shouldn't be done in PRs (here), otherwise LGTM. Could you remove the bump commit?

jbryson3:
Just reverted it, thanks for the guidance.

johanneswuerbach:
I created another PR without the two version bump and revert commits, going to merge when CI is green. :smile:

jbryson3:
/smackshead - should've thought to create a new PR instead of dirtying the history with a revert commit!

johanneswuerbach:
No worries, in the future you can also do an interactive rebase and force push to a PR.

PR #544, opened 2015-04-21
Use setRawMode just when tty is avaliable

marcioj:
I hit this problem while using testem as a subprocess in a integration test.
In the end of my test I send a `"q"` to testem stdin in order to quit the dev server, but it always fail with undefined is not a function error.


johanneswuerbach:
Restarted windows ci.

PR #548, opened 2015-04-24
In CI app, clean up launchers on SIGINT or SIGTERM

mzgoddard:
If SIGINT or SIGTERM is sent to the testem process in CI app, clean up
the launchers so that they aren't sitting around afterwords.

### DIFF
--- a/lib/ci/index.js
+++ b/lib/ci/index.js
@@ -54,6 +54,7 @@ App.prototype = {
   start: function(){
     log.info('Starting ci')
     async.series([
+      this.addSignalListeners.bind(this),
       this.startServer.bind(this),
       this.runHook.bind(this, 'on_start'),
       this.runHook.bind(this, 'before_tests'),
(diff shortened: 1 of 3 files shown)


mzgoddard:
Noticed on my system that --timeout times out the tests but then the process waits for any launched browsers to quit on their own or be killed by the normal completion of the tests. This adds a test and makes sure timeout leads to a quicker finish.

lib/ci/index.js
```diff
@@ -180,6 +191,47 @@ App.prototype = {
     if (this.exited) return
     this.cleanExit(this.getExitCode())
     this.exited = true
+  },
+
+  addSignalListeners: function(callback) {
+    this._boundSigInterrupt = function() {
+      this.wrapUp(new Error('Received SIGINT signal'))
+    }.bind(this)
+    process.on('SIGINT', this._boundSigInterrupt)
```
mzgoddard: This is the best place I've figured to put these so far. Happy to move them anywhere.

johanneswuerbach:
Thank you :+1:

PR #558, opened 2015-05-19
Force a single exe file for each browser once we found existing exe among other alternatives

vlizard:
Browser launcher can try to find one of several executables (e.g. [chromium on Linux](https://github.com/airportyh/testem/blob/master/lib/browser_launcher.js#L228)). In this case we can see reports like this in CI:
```
Chromium - Browser "chromium,chromium-browser --user-data-dir=/tmp/testem.chromium --no-default-browser-check --no-first-run --ignore-certificate-errors http://localhost:7357/3463" exited unexpectedly.
```
I believe, once we figure out which one of alternative exes exists we can fix it in settings for the rest of the current testem run.

### DIFF
--- a/tests/launcher_tests.js
+++ b/tests/launcher_tests.js
@@ -176,6 +176,17 @@ describe('Launcher', function(){
       assert.equal(launcher.commandLine(), '"node -e console.log(process.argv.slice(1).join(\' \')) hello"')
     })
 
+    it('returns commandLine with a single exe', function(done){
+      settings = {exe: ['node', 'npm'], args: function(){ return ['-e', 'console.log(1)'] }}
+      config = new Config()
+      launcher = new Launcher('single exe', settings, config)
+      launcher.launch()
+      launcher.on('processExit', function(){
+        assert.equal(launcher.commandLine(), '"node -e console.log(1)"')
+        done()
+      })
+    })
+
     it('copies the current environment', function(done) {
       var originalEnv = process.env;
       process.env.TESTEM_USER_CONFIG = 'copied'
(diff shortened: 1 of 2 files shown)


vlizard:
I'm not sure why AppVeyor reports that my test fails with timeout. Did I miss anything in the test itself or it's actually the code that is incorrect?

johanneswuerbach:
No, the overall PR looks good :+1:, I'll investigate why windows is complaining in a bit

tests/launcher_tests.js
```diff
@@ -170,6 +170,17 @@ describe('Launcher', function(){
       assert.equal(launcher.commandLine(), '"node -e console.log(process.argv.slice(1).join(\' \')) hello"')
     })
 
+    it('returns commandLine with a single exe', function(done){
+      settings = {exe: ['node', 'npm'], args: function(){ return ['-e', 'console.log(1)'] }}
```
johanneswuerbach: Could you remove the function? `args: ['-e', '1']` should be enough

vlizard: Removing function makes `commandLine` return something along `"node -e console.log(1) http://localhost:7357/1838"`. I didn't want to try matching URL and id as well so I ended up having function here.

johanneswuerbach:
I slightly refactored this area to remove the value leaking between tests. Lets see whether this passes: https://github.com/airportyh/testem/pull/559

vlizard:
I rebased PR against new master state. Let's check if it helped or not.

vlizard:
Nope, it does not. I think I'll have to find Windows machine to try it there.

johanneswuerbach:
Strange, I'll also take a look when I have access to my other machine.

vlizard:
Test passes on virtualbox'ed Windows 7 machine. Unfortunately I won't have time to check in other Windows versions until after weekend.

johanneswuerbach:
Thank you for this PR. I disabled the windows tests for now and merged your changes https://github.com/airportyh/testem/issues/566

vlizard:
Cool. Thank you.

Issue #561, opened 2015-05-19
Fail when explicitly defined launchers fail to start


johanneswuerbach:
A missing phantomjs should fail the ci run, when explicitly defined as `launcher_in_ci`.

Original issue: https://github.com/ember-cli/ember-cli/issues/3934

PR #565, opened 2015-05-27
Fail when explicitly defined

johanneswuerbach:
Fail when launchers explicitly defined in `launch_in_ci` are not
available.

Fixes https://github.com/airportyh/testem/issues/561


QuantumInformation:
Nice work thanks!

PR #570, opened 2015-06-02
Allow routes as arrays for file fallback

skeate:
I've found this handy. Use case, for me at least, was generating instrumented JS files, but having a fallback for (e.g.) vendor libraries:

```javascript
{
  //...
  "routes": {
    "/images": "public/images",
    "/js": ["instrumented", "public/javascripts"]
  },
  //...
}
```

Didn't write any automated tests (didn't see any for the `routes` property) but used it in another project and it seems to work a treat.

### DIFF
--- a/lib/server/index.js
+++ b/lib/server/index.js
@@ -276,7 +276,14 @@ Server.prototype = {
     for (var prefix in routes){
       if (uri.substring(0, prefix.length) === prefix){
         if (bestMatchLength < prefix.length){
-          bestMatch = routes[prefix] + '/' + uri.substring(prefix.length)
+          if (routes[prefix] instanceof Array) {
+            routes[prefix].some(function(folder) {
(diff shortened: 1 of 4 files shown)


johanneswuerbach:
Great addition, I'm just a little concerned about the missing tests. Could you add them by any chance? The server tests are here: https://github.com/airportyh/testem/blob/master/tests/server_tests.js

skeate:
Tests added. Incidentally they revealed a small bug; if the server isn't running in the same directory as what originally spawned it (I guess?) then the path it checks for existence is wrong.

Not sure why the Travis build failed; the test has nothing to do with my changes. It is checking the `tests/web/` folder, where I added some subfolders -- could that have changed the order of the files, for some reason?

johanneswuerbach:
Thank you, the CI failure was not related to your changes and is fixed on master.

PR #582, opened 2015-06-16
Preserve newlines in YAMLish of TAP reports

### LINKED ISSUE #581: Log lines folded in TAP diagnostics
TAP produces output a la:

```
ok 1 PhantomJS 2.0 - Check output
    ---
        Log: >
            first log line
            second log line
            third log line
    ...
```

Due to the usage of the folding style (denoted by the `>` ), the log output gets folded into a single line when read by a TAP consumer (ie. when displayed in Jenkins).

I propose to change that to use the [literal style](http://www.yaml.org/spec/1.2/spec.html#id2760844) (denoted by `|`) to preserve the line breaks.

What do you think? Should I submit a PR?

avdv:
Use `|` instead of `>` to denote scalar content written in literal style
in "Log" diagnostics when producing TAP output.

Fixes #581.

### DIFF
--- a/lib/ci/test_reporters/tap_reporter.js
+++ b/lib/ci/test_reporters/tap_reporter.js
@@ -30,7 +30,7 @@ TapReporter.prototype = {
         return key + ': >\n' + strutils.indent(String(err[key]))
       })
     if(logs){
-        var testLogs = ["Log: >"].concat(logs.map(function(log){return strutils.indent(String(log))}))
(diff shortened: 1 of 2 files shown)


johanneswuerbach:
Hey, sorry for the late reply. Could you rebase this commit and check whether this makes our windows CI pass?

avdv:
> Hey, sorry for the late reply. Could you rebase this commit and check whether this makes our windows CI pass?

No problem. Here goes.

Issue #392, opened 2014-07-30
Officially sanctioned way to use a custom reporter


miquella:
We have an analytics tool called Firework that we're trying to integrate with testem, but we need a [custom reporter](https://github.com/miquella/testem-firework-reporter) to do so.

I've [hacked](https://github.com/miquella/testem-firework-reporter/blob/9f7550cc958a83117841c49078aba3e5bfd6dada/index.js#L38-L39) around the problem for now by directly injecting my reporter into the testem's [test_reporters](https://github.com/airportyh/testem/blob/master/lib/ci/test_reporters/index.js), but I'm hoping for an officially sanctioned way of using a custom reporter. Is this a possibility? Have I missed a decent way of doing this that already exists?

mieky:
It would be neat to enable the use of Mocha reporters (as [pointed out](https://github.com/airportyh/testem/issues/113#issuecomment-10988719) before). For example, the xunit reporter only prints out its XML once finished, while https://github.com/juhovh/mocha-jenkins-reporter supports printing both to the console and an output file.

PR #597, opened 2015-07-07
add docs for writing and configuring a custom reporter

kielni:
I wanted to write a custom reporter.  I found what I needed from a combination of https://github.com/airportyh/testem/issues/392 and the code for the built-in reporters.  It'd be nice if it were easier to find, so I updated the docs.

For reference, my custom reporter is https://github.com/Gridium/testem-failure-reporter


airportyh:
That's fantastic! Thanks @kielni!

PR #605, opened 2015-07-18
Add qunit logs to xunit ci

BryanCrotaz:
Pass qunit logs and messages into xunit output node


johanneswuerbach:
Could you add a test for your change?

PR #617, opened 2015-07-30
Using custom browser id if provided

paulot:
Currently I'm working on a wrapper for testem that will enable screenshots of the test-page (particularly useful for for ember applications). For that, the wrapper needs to launch a custom browser via selenium. Since, launching selenium browsers from testem itself is not desirable, we need to instrument the selenium browser from outside of testem. The problem with it is that there is no way to find the id that was assigned to the custom browser from outside of testem. This change addresses that, by allowing the user to specify the id to be used with the custom browser in the config file.

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -12,7 +12,7 @@ function Launcher(name, settings, config){
   this.config = config
   this.settings = settings
   this.setupDefaultSettings()
-  this.id = String(Math.floor(Math.random() * 10000))
+  this.id = settings.id || String(Math.floor(Math.random() * 10000))
 }
 
 Launcher.prototype = {


johanneswuerbach:
Could you add an example how such config would look like?

paulot:
Sure,
Inside in the test command for ember cli in the test command (see https://github.com/ember-cli/ember-cli/blob/master/lib/commands/test.js for context) we could have something like:
``` javascript
_generateCustomConfigFile: function(options) {
  55-68 unchanged
  originalContents['launchers'].push({
    selenium_chrome: {  
      command: 'noop',
      id: 10001,
      protocol: 'browser'
    }
  });
  fs.writeFileSync(customPath, JSON.stringify(originalContents));

  return customPath;
},
```
Then in run we can have something like:
``` javascript
run: function(commandOptions) {
  if ('selenium_chrome' in commandOptions.launch) {
    // Launch the selenium webdriver for chrome using the id assigned above
  }
  ...
}
```
So, this would give a lot more autonomy to ember plugins (or just about anything built on top of testem) as it would allow the wrapper to instrument the browser themselves. This is particularly useful if we want to instrument the browser with selenium to take screenshots. So the config modification that I mean really would not happen at the user level, but at the level of a wrapper, like ember test.

johanneswuerbach:
Thanks, SGTM :+1:

PR #619, opened 2015-08-04
Enable XUnit to report intermediate test results.

davidhao3300:
Enable the XUnit reporter to display test suite progress like the TAP reporter does. By default it chooses process.stdout unless if process.stdout is already taken, in which case it uses process.stderr.

We wanted this feature because running tests on Jenkins provides little context on how a test is doing until the very end.

Most of the changes are just moving shared code into a separate file.

### DIFF
--- a/lib/ci/test_reporters/xunit_reporter.js
+++ b/lib/ci/test_reporters/xunit_reporter.js
@@ -1,9 +1,10 @@
-var strutils = require('../../strutils')
+var displayutils = require('../../displayutils')
 var XmlDom = require('xmldom')
 
 
 function XUnitReporter(silent, out){
   this.out = out || process.stdout
+  this.intermediateOut = this.out == process.stdout ? null : process.stdout
   this.silent = silent
   this.stoppedOnError = null
   this.id = 1
@@ -19,6 +20,7 @@ XUnitReporter.prototype = {
       launcher: prefix,
       result: data
     })
+    this.display(prefix, data)
     this.total++
     if (data.passed) this.pass++
   },
@@ -44,6 +46,10 @@ XUnitReporter.prototype = {
     }
 
     return new XmlDom.XMLSerializer().serializeToString(doc.documentElement);
+  },
+  display: function(prefix, result){
+    if (this.silent || !this.intermediateOut) return
+    this.intermediateOut.write(displayutils.resultString(this.id++, prefix, result))
   },
   getTestResultNode: function(document, result){
     var launcher = result.launcher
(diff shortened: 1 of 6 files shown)


johanneswuerbach:
Thanks, this looks good. Could you add tests and docs?

davidhao3300:
Sure

davidhao3300:
And done. I also added documentation about configuring report_file, since that wasn't something I was aware of before looking into the source code.

lib/ci/test_reporters/xunit_reporter.js
```diff
@@ -1,9 +1,10 @@
-var strutils = require('../../strutils')
+var displayutils = require('../../displayutils')
 var XmlDom = require('xmldom')
 
 
 function XUnitReporter(silent, out){
   this.out = out || process.stdout
+  this.intermediateOut = this.out == process.stdout ? process.stderr : process.stdout
```
johanneswuerbach: Could we disable the `intermediateOut` when stdout is already used? I don't like the idea printing to `stderr` when everything is fine.

davidhao3300: Yeah, that makes sense.

davidhao3300:
Ready for review again, rebased on master and squashed.

johanneswuerbach:
Thanks :-)

johanneswuerbach:
Hm, looks like something didn't work as expected, I'm seeing both now:
```
ok 1 PhantomJS 2.0 - says hello world
ok 2 PhantomJS 2.0 - says hello to person
<testsuite name="Testem Tests" tests="2" failures="0" timestamp="Thu Aug 06 2015 22:07:53 GMT+0200 (CEST)" time="0.821"><testcase classname="PhantomJS 2.0" name="says hello world" time="0.001"/><testcase classname="PhantomJS 2.0" name="says hello to person" time="0"/></testsuite>
```

with 
```
{
    "framework": "qunit",
    "reporter": "xunit",
    "report_file": "test.xml"
}
```

davidhao3300:
I'll look into this.

davidhao3300:
In lib/ci/report_file.js, the ReportFile write both to the file and to stdout. Not sure if we want to modify this behavior.

johanneswuerbach:
Sorry, I'll revert this a the mixed output may be confusing and breaking others.
https://github.com/airportyh/testem/pull/623

I generally like the idea of having different reporters on file and stdout, but then they shouldn't mix produce unparseable mixed output.

Looks like a rspec like formatter chain would be my dream, but might be complex to implement.
http://www.relishapp.com/rspec/rspec-core/v/3-3/docs/command-line/format-option
When chain support would be implemented your config could look like:
`--format tape --format xunit --out testem-xunit.xml` where xunit is written and tape is printed to stdout.

PR #621, opened 2015-08-06
Adds config option to attach event listeners to client Testem socket

joseph-chu:
Can be used with custom reporters to send extra information from the browser.

```
browser_event_listeners: [
  {
    eventName: 'special-browser-event'
    eventListener: function(specialEventData) {
      console.log("This event was probably sent through window.Testem.emit");
      console.log("The browser that sent it was: " + this.browserName);
      console.log(specialEventData);
    }
  }
]
```

### DIFF
--- a/docs/config_file.md
+++ b/docs/config_file.md
@@ -77,12 +77,12 @@ Common Configuration Options
     cwd:                      [Path]    directory to use as root
     parallel:                 [Number]  max number of parallel runners (1)
     routes:                   [Object]  overrides for assets paths
-    fail_on_zero_tests:       [Boolean] whether process should exit with error status when no tests found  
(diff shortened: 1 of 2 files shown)


johanneswuerbach:
Great, could you add some tests?

joseph-chu:
I'll work on that.  It looks like it doesn't fit nearly into the existing `ci` tests so I'll probably create a new file for it.

johanneswuerbach:
Thanks, I think the easiest way is to create another test fixture, which emits events and model a test like this:
https://github.com/airportyh/testem/blob/master/tests/ci/ci_tests.js#L39-L44

Also this needs rebasing now.

PR #622, opened 2015-08-06
Show Error Messages correctly in dev

paytonrules:
The error messages in the before_tests and onStart hooks relied on the
error object having stdout and stderr properties. Unfortunately they
don't - they use the Error object which only has a name and message
property. This meant that if one of those hooks failed the error would
say:

Error
undefined
undefined

I fixed this and removed a little duplication. There are holes here in
the testing, which I didn't fix.

See
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error
for a description of the JS standard Error object.

### DIFF
--- a/lib/dev/index.js
+++ b/lib/dev/index.js
@@ -75,7 +75,7 @@ function App(config, finalizer, cb){
 App.prototype = {
   __proto__: EventEmitter.prototype,
   start: function(){},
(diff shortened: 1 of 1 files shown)


johanneswuerbach:
Thanks, I ll look into adding a test for this

PR #625, opened 2015-08-12
Xunit intermediate output v2.

davidhao3300:
This is similar to the previous PR, except that it will not affect any existing behavior. Adds a flag to enable xunit intermediate output (false by default). If people turn on the flag, it shows that they do indeed desire/expect intermediate output.


davidhao3300:
ping on this

johanneswuerbach:
Sorry, lgtm :-)

davidhao3300:
Thanks!

PR #644, opened 2015-10-06
Add query_params config

asakusuma:
Allow consumers to add query params to the test page. This allows your
config to set the test page url and a programatic API call to set the
query params.


johanneswuerbach:
Thanks, published `v0.9.7`

PR #649, opened 2015-10-15
Add PhantomJS for FreeBSD

rzurad:
Hi there,

I'm one of the few folks whose development environment is FreeBSD 10, which makes running Ember-CLI tests through testem difficult because testem has no launchers for FreeBSD.

This commit adds PhantomJS as a default launcher for FreeBSD. I've tested this through Ember-CLI v1.13.8 and FreeBSD 10.2 with the ports package for PhantomJS 2.0.0 and it seems to work just fine.

Let me know if more is needed for this

Thanks!

### DIFF
--- a/lib/browser_launcher.js
+++ b/lib/browser_launcher.js
@@ -253,6 +253,15 @@ function browsersForPlatform(){
           supported: findableByWhichOrModule
         }
   	]
+  }else if (platform === 'freebsd') {
+  	return [
+        {
+          name: 'PhantomJS',
+          exe: 'phantomjs',
+          args: buildPhantomJsArgs,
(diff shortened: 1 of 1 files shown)


johanneswuerbach:
Thank you for adding this!

rzurad:
My pleasure :)

Any ETA on when a tag is coming out that will include this merge? I'd like to submit a pull request to Ember-CLI as soon as possible that updates the version of testem that they're using to one that includes this merge. I'm betting they'd prefer to use a version number rather than a commit hash in their package.json file.

Thanks again!

johanneswuerbach:
Sure, released `0.9.8`

rzurad:
Thanks! :)

PR #653, opened 2015-10-30
Node 5

johanneswuerbach:


johanneswuerbach:
Node 0.10 and 0.12 looks unrelated, but need to be debugged.

PR #664, opened 2015-11-15
Merged modes

johanneswuerbach:


airportyh:
Does this mean there no longer a `testem ci`? I am curious how this will work. Also, documentation will need to change - which I am willing to help out with.

johanneswuerbach:
No, all commands etc. will be the same. Code was just merged and the only difference between dev and ci is now some settings: https://github.com/testem/testem/blob/master/lib/config.js#L34-L42

johanneswuerbach:
This was mostly done to have one place for all logic (hooks, tests, runners etc.) and to maybe add more modes in the future, but this still has to be defined.

PR #668, opened 2015-11-15
Converted sources to use more prevalent JavaScipt coding style

airportyh:
In particular, these styles were enforced:
    
    * always use semicolons
    * always add a space before the beginning curly a block and after the closing curly of a block
FYI, I wrote this code rewriter to do the transform https://gist.github.com/airportyh/996c1f1afe7105e3eb07


johanneswuerbach:
:+1: amazing

PR #606, opened 2015-07-19
Output skipped tests from qunit to xunit

BryanCrotaz:
Outputs skipped tests with <skipped/> tag

PR #674, opened 2015-11-17
Output skipped tests from qunit to xunit & tap

step2yeung:
This is a copy of #606 with the additional reporting in tap reporter.
Since #606 has not been touched in 4 months, I created a new PR.

This will output skipped tests for both xunit & tap reporter.

Fixes #606, fixes https://github.com/testem/testem/issues/611


step2yeung:
@nickiaconis

johanneswuerbach:
Thanks, LGTM.

PR #677, opened 2015-11-17
Having skipped tests is ok

nickiaconis:
cc @step2yeung


step2yeung:
Thanks Nick!

johanneswuerbach:
Thanks, could one of you add a test for this? Seems important to not regress here.

nickiaconis:
Good point. I've updated the tap reporter tests to include a case without failures.

johanneswuerbach:
Thanks, :bow:

PR #678, opened 2015-11-18
Don't count skipped tests as passed

nickiaconis:


johanneswuerbach:
Hey, could you also add a test for that? Thanks :bow:

nickiaconis:
Certainly! I didn't see any existing place where the new tests seemed to fit, so I created a new file. Let me know if there's somewhere else you'd prefer them.

johanneswuerbach:
thanks, looks great!

PR #680, opened 2015-11-19
Use latest Jasmine2 version.

zigomir:
This is just a proposal of how to handle future Jasmine2 upgrades.

I know this is a compromise because to run example you now need to exactly specify `src_files` which before probably was not needed. How can we do it this way without a compromise?


johanneswuerbach:
Thanks for starting this. I was already playing around with different solutions and was favouring loading the test frameworks via a CDN like http://cdnjs.com/libraries/jasmine by default and allow developers to bring their own local version. This would still allow a quick start, but decreases package size and removes the maintenance burden. Most devs I have talked to are already overwriting the bundled version with their specific version fetched via npm / bower. What do you think?

zigomir:
I'm totally for some kind of customization of testing libs. But I couldn't find a way to do it so I came up with this. Can you give me more direction on how it's usually done when you want something like this?

Maybe I can add something to docs instead. Thanks!

airportyh:
I have given this some thought previously a well. I also tried to go with the npm module approach, and I tried to model each testing framework as a plugin/peer dependency. I.e. `npm install testem-jasmine`. I sketched it out at https://github.com/airportyh/testem2 but alas that effort was abandoned.

I worry a little bit about using CDN by default because that prevents your tests from running offline.

johanneswuerbach:
@zigomir you can overwrite the jasmine version by adding your own `jasmine2-boot.js`, `jasmine2-html.js` and `jasmine2.js` to `srv_files`.

zigomir:
@airportyh yeah, offline abillity is a must imo.

@johanneswuerbach that's great, thank you!

so I'm probabl going to close this PR since you already considered these options. I'll go through readme again and maybe propose a section how to get latest testing framework in. Is thay OK?

johanneswuerbach:
Yes, that sounds great :+1:

PR #686, opened 2015-11-21
Use css_files in custom runner.

zigomir:
This is kinda related to https://github.com/testem/testem/pull/680. Because if we want to use latest Jasmine you also want to include the CSS.

With that in place, one can easily write this kind of config file and get latest Jasmine:

```json
{
    "framework": "custom",
    "src_files": [
        "node_modules/jasmine-core/lib/jasmine-core/jasmine.js",
        "node_modules/jasmine-core/lib/jasmine-core/jasmine-html.js",
        "node_modules/jasmine-core/lib/jasmine-core/boot.js",
        "dist/test.js"
    ],
    "css_files": [
        "node_modules/jasmine-core/lib/jasmine-core/jasmine.css"
    ],
    "launch_in_dev": [ "chrome", "node" ]
}
```

I'll add it to README later on, if this PR makes sense.

### DIFF
--- a/views/customrunner.mustache
+++ b/views/customrunner.mustache
@@ -4,6 +4,7 @@
   <title>Test'em</title>
   <script src="/testem.js"></script>
   {{#scripts}}<script src="{{{src}}}"{{#attrs}} {{&.}}{{/attrs}}></script>{{/scripts}}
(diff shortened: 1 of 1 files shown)


johanneswuerbach:
Thanks, :+1:

PR #689, opened 2015-11-24
Wrap Linux browser launches with dbus-launch when possible

davidhao3300:
I noticed that Chrome sometimes fails to start properly on Linux VMs. Recently, a solution has been found here: https://github.com/angular/protractor/issues/2419 that solves the issue (I've confirmed that by running tests repeatedly and getting no Chrome hangs). I implemented their solution here. This first checks whether the requisite executable is installed before wrapping the browsers, so this shouldn't cause any regressions.

### DIFF
--- a/lib/browser_launcher.js
+++ b/lib/browser_launcher.js
@@ -59,11 +59,11 @@ function buildPhantomJsArgs(config) {
 // * `setup(app, done)` - any initial setup needed before launching the executable(this is async,
 //        the second parameter `done()` must be invoked when done).
 // * `supported(cb)` - an async function which tells us whether the browser is supported by the current machine.
-function browsersForPlatform() {
+function browsersForPlatform(cb) {
   var platform = process.platform;
 
   if (platform === 'win32') {
-    return [
+    cb([
       {
         name: 'IE',
         exe: ':\\Program Files\\Internet Explorer\\iexplore.exe',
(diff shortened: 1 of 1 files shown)


davidhao3300:
not sure how to test this

johanneswuerbach:
Thanks :+1:, looks like your change broke CI mind taking a look?

davidhao3300:
yeah, I'll get around to it today.

davidhao3300:
Okay, all checks have passed

lib/browser_launcher.js
```diff
@@ -239,6 +240,19 @@ function browsersForPlatform() {
         supported: findableByWhichOrModule
       }
     ];
+
+    // If dbus-launch is present, wrap google-chrome launch
+    // to prevent a race condition in Google Chrome
+    // Discussion here: https://github.com/angular/protractor/issues/2419
+    if (whereis('dbus-launch')) {
```
johanneswuerbach: We already have a function, which does this. Could you replace `node-whereis` with https://github.com/testem/testem/blob/master/lib/fileutils.js#L19?

davidhao3300: sure

davidhao3300:
Alright, changes have been made. I noticed there was a case where dbus-launch could be installed but not the original browser. This now handles that by first checking the availability of the original browser before wrapping the browser in dbus-launch.

johanneswuerbach:
Thanks :+1:

Issue #171, opened 2013-02-22
Display error if on_start errors out


svnlto:
Browserstack expects you to have a config file with credentials under your ~ 
Testem should error out if the file does not exists but it's config contains

```
#on_start:
  #command: browserstack tunnel localhost:<port

```

LayZeeDK:
#690 is merged and should fix this issue, so it should be closed.

PR #690, opened 2015-11-27
Testem - BrowserStack examples

vedharish:
Working examples for using Testem with BrowserStack.

The issue was that the old examples used [browserstack-cli](https://github.com/dbrans/browserstack-cli) which is quite old (last commit was in 2013).

These examples use instead [browserstacktunnel-wrapper](https://github.com/pghalliday/node-BrowserStackTunnel) for setting up the local tunnel (we wont need to download the local binary and set it up ourselves) and [browserstack](https://github.com/scottgonzalez/node-browserstack) as a BrowserStack client to run the tests.


vedharish:
Closes #171 (no longer expects the config file)

johanneswuerbach:
Thanks :+1:

PR #693, opened 2015-12-03
Allow custom user data directories

davidhao3300:
This allows users to specify a directory for the browser user data profiles. Default behavior is the same as before.

Not sure how to test this.


johanneswuerbach:
Thanks :+1: could you add a test for that?

davidhao3300:
Okay this is done

johanneswuerbach:
Thanks again, looks great :+1:

PR #698, opened 2015-12-08
Prevent multiple calls to callback error

davidhao3300:
This is a regression caused by https://github.com/testem/testem/commit/79e75d850942804e197d4bb676b9cbc5c4026b53

It appears that removing the `socket.once` allows multiple calls to the same callback, which is bad.


johanneswuerbach:
Thanks, I'm currently in thr process of adding more tests to the runners, hopefully that will prevent further regressions.

PR #702, opened 2015-12-11
Use mkdirp instead of mkdir to handle creating parent directories

davidhao3300:
Because we allow custom data directories, we now face an issue where we attempt to create a directory whose parent directory hasn't been created yet. Use mkdirp to handle creating parent directories.
If fs.mkdir = mkdir, then mkdirp = mkdir -p


johanneswuerbach:
Thanks, could you add a test? Seems like an easy regression.

davidhao3300:
done

tests/config_tests.js
```diff
@@ -246,14 +247,19 @@ describe('Config', function() {
...
-      var correctFlag = config.cwd() + '/customDirectory/testem.firefox';
+      var correctFlag = config.cwd() + '/node_modules/testem.firefox';
       if (process.platform === 'win32') {
-        correctFlag = config.cwd() + '\\customDirectory\\testem.firefox';
+        correctFlag = config.cwd() + '\\node_modules\\testem.firefox';
       }
       expect(launchers.firefox.settings.args).to.include(correctFlag);
-      done();
+      launchers.firefox.settings.setup(config, function() {
+        fs.exists(correctFlag, function(exists) {
```
johanneswuerbach: Hm, `fs.exists` is deprecated https://nodejs.org/api/fs.html#fs_fs_exists_path_callback, could you use https://nodejs.org/api/fs.html#fs_fs_stat_path_callback instead?

johanneswuerbach:
small nit, otherwise :+1:

davidhao3300:
done

PR #706, opened 2015-12-17
Removes embedded frameworks

johanneswuerbach:
Embedded frameworks were already really outdated and developers can easily
ship their own libraries. Using the default framework template will now
load the test framework and styles from the internet. This still allows an easy start, but stops shipping framework code twice we you use your own version anyway.


johanneswuerbach:
//cc @airportyh

jasonkarns:
Any chance of getting some improved support for loading from node_modules? Maybe just specifying the module name and testem expands to `node_modules/<module>/<main>` ?

johanneswuerbach:
Sure, mind creating a separate issue for that? With all transpiling and es6 modules recently I don't see a really strong case for it, but PRs are definitely welcome.

jasonkarns:
In prepping the issue, I checked to see if either mocha or jasmine's `main` file was the browser-compatible version (neither are) or if they expose the browser build via `browser` (they do not).  Until npm's browser script support is more robust and browser entrypoints are specified via package.json, this feature wouldn't save much as explicit paths to the browser file would still be required. Assuming test frameworks eventually expose `browser` or similar to point to the browser build, then this feature would be beneficial.

searls:
I just realized that I've been dealing with intermittent build failures for weeks, and the root cause was that I can't run my tests when I don't have an Internet connection.

I'm frequently without an Internet connection, so this change makes Testem much less usable for me. I understand the motives for this PR, but what can we do to make sure testem is still usable without an Internet connection?

johanneswuerbach:
You can fetch the test framework you are using manually or via npm [example](https://github.com/testem/testem/tree/master/examples/jasmine_custom), this also allows you to fully control the version of the test framework.

PR #709, opened 2015-12-22
Asynchronously detect + setup the testFramework

stefanpenner:
allows:

* testem.js to be loaded as soon as possible
* allowing testem to catch all errors
  * as the test framework itself can fail to load
* much better user experience in failure scenarios

### DIFF
--- a/public/testem/testem_client.js
+++ b/public/testem/testem_client.js
@@ -75,10 +75,13 @@ function initTestFrameworkHooks(socket) {
 }
 
 function init() {
-  takeOverConsole();
   interceptWindowOnError();
-  initTestFrameworkHooks(Testem);
+  takeOverConsole();
   setupTestStats();
+  Testem.initTestFrameworkHooks = function() {
+    initTestFrameworkHooks(Testem);
+  };
+  initTestFrameworkHooks(Testem);
 }
 
 function setupTestStats() {
@@ -152,6 +155,8 @@ function emit() {
 }
 
 window.Testem = {
+  // set during init
+  initTestFrameworkHooks: undefined,
   emitConnectionQueue: [],
   useCustomAdapter: function(adapter) {
     adapter(this);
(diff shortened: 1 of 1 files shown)


stefanpenner:
well, this ended up being just as racy as I thought... :P

stefanpenner:
Ok, took a different approach. Now this exposes `Testem.initTestFrameworkHooks` which can be called once the test framework is loaded.

johanneswuerbach:
Looks like a good idea, ideas how we could make this not a breaking change?

stefanpenner:
@johanneswuerbach it should already be, existing semantics remain. One can now just re-invoke `initTestFrameworkHooks`

stefanpenner:
I suspect the only thing we should maybe do, is track if `initTestFrameworkHooks` did attach, and if it did blackhole future calls to `initTestFrameworkHooks`.

johanneswuerbach:
Ah now I got it. Yes black-holing sounds good I would prefer to throw an error if called manually an unable to attach. Otherwise this would silently fail and just report nothing.

At this point everything should be instrumented so just throwing or console.error should be fine.

stefanpenner:
@johanneswuerbach https://github.com/testem/testem/commit/d2a427433c4377c5bbacca5414347f36361b5864 should likely prevent the issue.

This whole thing likely needs a test, but I need to sleep. It should be good to go, unless you have further thoughts. If no-one beats me to it, I can likely write some tests tomorrow afternoon.

johanneswuerbach:
@stefanpenner, thanks I can take it from here. This whole code part is sadly just covered by integration-tests, but I'll check what I can do.

stefanpenner:
@johanneswuerbach i added a quick assertion if the manual call fails (as per your suggestion), i suspect the message should be given some more TLC

johanneswuerbach:
Did some cleanup and added an integration test https://github.com/testem/testem/issues/710

PR #710, opened 2015-12-22
Testem.hookIntoTestFramework to manually hook into your test framework

johanneswuerbach:
Exposed Testem.hookIntoTestFramework do allow hooking manually into a
test framework. This allows loading `testem.js` as first script to
catch parsing issues when framework and tests are bundled.

Replaces https://github.com/testem/testem/pull/709

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -156,7 +156,8 @@ Config.prototype.defaults = {
     return scheme + '://' + this.get('host') + ':' + this.get('port') + '/';
   },
   parallel: 1,
-  reporter: 'tap'
+  reporter: 'tap',
+  bail_on_uncaught_error: true
 };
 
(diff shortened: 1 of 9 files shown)


lib/config.js
```diff
@@ -156,7 +156,8 @@ Config.prototype.defaults = {
     return scheme + '://' + this.get('host') + ':' + this.get('port') + '/';
   },
   parallel: 1,
-  reporter: 'tap'
+  reporter: 'tap',
+  bail_on_uncaught_error: true
```
johanneswuerbach: While this is technically a breaking change, I think its a better default and I'm currently preparing a `1.0` so small breaking changes should be fine.

stefanpenner:
LGTM

PR #722, opened 2016-01-04
adding test for POST proxy request

quantumlicht:


quantumlicht:
I only provided tests for now. If someone can point me in the general direction of where the fix might be, I'd be happy to assist in fixing it.

johanneswuerbach:
Proxy setup is done here, https://github.com/testem/testem/blob/master/lib/server/index.js#L175-L197 We delegate all the hard work to https://github.com/nodejitsu/node-http-proxy, so we either misuse that module or it might have the same issue.

quantumlicht:
They currently have this issue opened: https://github.com/nodejitsu/node-http-proxy/issues/180

PR #728, opened 2016-01-07
Revert to official socket.io

johanneswuerbach:
v1.4.0 no longer ships with ws with native dependencies


johanneswuerbach:
Looks like this needs to wait for `1.4.1`, because of https://github.com/socketio/socket.io-client/issues/930

PR #742, opened 2016-01-21
Pass on associated stacktrace from qunit log callback

kkumler:
The [QUnit.log callback](http://api.qunitjs.com/QUnit.log/) will potentially pass a stacktrace. This should pass it along if it is available.

NB: I did not see a facility to test the QUnit adapter.

### DIFF
--- a/public/testem/qunit_adapter.js
+++ b/public/testem/qunit_adapter.js
@@ -55,7 +55,7 @@ function qunitAdapter() {
         passed: params.result,
         line: lineNumber(e),
         file: sourceFile(e),
-        stack: stacktrace(e),
+        stack: stacktrace(e) || params.source,
         message: message(e)
       });
     } else {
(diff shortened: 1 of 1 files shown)


public/testem/qunit_adapter.js
```diff
@@ -55,7 +55,7 @@ function qunitAdapter() {
         passed: params.result,
         line: lineNumber(e),
         file: sourceFile(e),
-        stack: stacktrace(e),
+        stack: stacktrace(e) || result.source,
```
johanneswuerbach: Looks like `result` is never defined.

johanneswuerbach:
Yes, sadly this area is only integration test covered and not unit tested.

public/testem/qunit_adapter.js
kkumler: Oops, fixed.

PR #743, opened 2016-01-21
Fix dot test reporter outputting F for passed tests

amk221:
the dot test reporter was outputting F's `.F.F.F.F.F.F` regardless


johanneswuerbach:
:+1:

PR #748, opened 2016-01-25
Made unit tests and integration tests to run out of the box (and in Travis)

alexindigo:
I think it should help onboarding if all you need as a new developer on the project, is to run:

```
$ npm install
$ npm test
$ npm run integration
```

At the moment it requires chasing different dependencies, installing them manually,
and executing different scripts depending on the platform of choice.

Also this PR updates some of the examples to make them work at all and/or to have up-to-date dependencies.

PS. Still going through windows-incompatible example, to make it pass on appveyor.


examples/node_example/testem.json
```diff
@@ -1,9 +1,10 @@
...
-        "Node": {
-            "command": "mocha tests.js"
-        }
-    },
-    "launch_in_dev": [ "Node" ],
-    "launch_in_ci": [ "Node" ]
+  "launchers": {
+    "Node": {
+      "command": "mocha tests.js -R tap",
+      "protocol": "tap"
```
johanneswuerbach: The `node_tap_example` should cover that case already, mind removing the proto and the `-R` tap to ensure only the process exit code is verified.

package.json
```diff
@@ -71,8 +73,10 @@
     "jscs": "^2.8.0",
     "jshint": "^2.8.0",
     "mocha": "^2.1.0",
+    "phantomjs": "^1.9.19",
```
johanneswuerbach: Right now this is excluded on purpose to be able to test different phantomjs versions. I would prefer to leave it like this.

alexindigo: This package checks for the existing installation and picks it up if present. So it won't break your process.

This is installation log from Travis:

```
Found PhantomJS at /usr/local/phantomjs/bin/phantomjs ...verifying
Writing location.js file
PhantomJS is already installed on PATH at /usr/local/phantomjs/bin/phantomjs
```

examples/node_example/testem.json
alexindigo: Removed it. I wasn't sure if broken output was intended or just something got mangled along the way.

package.json
johanneswuerbach: Yes, but it will only use the phnatomjs in your PATH, when it has the same version: https://github.com/Medium/phantomjs/blob/master/install.js#L466-L484

E.g. when we install phantomjs 2.0 manually on appveyor, this would overwrite it and always use 1.9.

alexindigo: So if we add extra step to appveyor, rather than requiring humans to do extra step (there is no mentioning of it in the readme) would it work for you? I'll adjust my PR to illustrate the idea.

PS. Right now my PR is broken because of the change in `jscs` they released something new (in their dep-tree) and existing version `2.8.0` picks it up, unmodified files have issues now.

johanneswuerbach: > PS. Right now my PR is broken because of the change in jscs they released something new (in their dep-tree) and existing version 2.8.0 picks it up, unmodified files have issues now.

I can take care of that.

> So if we add extra step to appveyor, rather than requiring humans to do extra step (there is no mentioning of it in the readme) would it work for you? I'll adjust my PR to illustrate the idea.

Won't really work as it would require, manually removing that `devDependency` on appveyor before running. I would prefer to just state that tests need phantomjs installed and add that you could install it using `npm install -g phantomjs` or `brew install phantomjs`.

johanneswuerbach:
Master should be passing again :smile:

alexindigo:
@johanneswuerbach Ok, I removed phantomjs, cleaned up other things. And it's passing travis and appveyour.

BTW, looks like your travis is running on OSX only now. :)

bin/run-integration.js
```diff
@@ -0,0 +1,106 @@
...
+  process.exit(0);
+}
+
+// get extra params
+var argv = process.argv.slice(2);
+var testFlags = argv.length ? ' ' + argv.join(' ') : '';
+
+// use custom testCmd instead of `npm test`
+// while we continue to support node@0.10 with npm@1
+var testCmd = 'npm run test' + (testFlags ? ' -- ' + testFlags : '');
```
johanneswuerbach: Don't we skip npm < 10 already above?

CONTRIBUTING.md
```diff
@@ -3,6 +3,16 @@ Contributing to Testem
...
 Open source is all about DIY! If you want something fixed, it's sometimes faster to just roll your sleeves up, not to mention lots more rewarding. This doc will give you some pointers on where to look when you poke around Testem's source code.
 
+Getting Started
+---------------
+
+* Fork and checkout [github.com/testem/testem](https://github.com/testem/testem)
+* Run `npm install` and `npm test` to make sure you're off to a good start
+
+> Testem expects `PhantomJS` to be in the PATH, if you don't have one,
+> either install it globally via `sudo npm install -g phantomjs`
```
johanneswuerbach: Could you remove the `sudo` here? Reasoning: https://github.com/testem/testem/pull/598#issuecomment-121334629

examples/browserify/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
johanneswuerbach: Could you use `/node_modules`? Otherwise this matches any kind of node modules folder, which might not be what you want.

johanneswuerbach:
Some small nits otherwise LGTM. :+1: 

> BTW, looks like your travis is running on OSX only now. :)

No, travis is just somehow rendering some options weird inside their UI.

johanneswuerbach:
Btw. have you tried running those examples in parallel? 10mins added test time on windows is a bit much :-(

alexindigo:
I could try to make them in parallel. Although tests would definitely die on my laptop. :) I have old (and struggling) Air. :)

CONTRIBUTING.md
alexindigo: Here, I'm with @avtar ob that, I now only handful of people who are using nvm. Most of the devs around me just got package from nodejs.org.
So I do think it's a default behavior.

That said, I'm willing to rephrase to accommodate both cases.

bin/run-integration.js
alexindigo: Left over comment from previous implementation. :) Good catch.

examples/browserify/.gitignore
alexindigo: This .gitignore affects only browserify example folder, so it does touch anything on the outside. Or there is something else you're thinking about?

alexindigo:
As for travis/osx, check ruby's version output.

CONTRIBUTING.md
johanneswuerbach: :+1:

examples/browserify/.gitignore
johanneswuerbach: I was more thinking that people will just copy those examples, but this won't work anyway, so feel free to leave it like that :-)

johanneswuerbach:
> I could try to make them in parallel. Although tests would definitely die on my laptop. :) I have old (and struggling) Air. :)

Will merge for now and look into decreasing the runtime in the next step.

> As for travis/osx, check ruby's version output.

Not sure what you mean:
```
$ ruby --version
ruby 1.9.3p551 (2014-11-13 revision 48407) [x86_64-linux]
```

```
$ ruby --version
ruby 2.0.0p643 (2015-02-25 revision 49749) [x86_64-darwin15.0.0]
```

Looks like travis is correctly using Ubuntu / OS X

alexindigo:
Heh, checked it again and you're right. I must've clicked on the same link few times. :) So much for working on PRs at night :)

bin/run-integration.js
alexindigo: Cleaned up.

CONTRIBUTING.md
alexindigo: Updated.

alexindigo:
I gave it a try (we already have `async` as dependency and it's capable enough of handling parallel tasks. There is different problem though :) for now all the examples run testem on the same port :)
would need to update them not to collide on ports, and/or make them run on random ports.

johanneswuerbach:
I would keep the PR like this and add parallel running in a separate one when time allows it. Mind squashing your commits? Then we could merge from my side

alexindigo:
Squashed commits. Let see if travis is happy. :)

alexindigo:
@johanneswuerbach Installation failed in appveyor, do you mind to restart the test? Thanks.

PR #750, opened 2016-01-25
Fix filter predicate for error items

avdv:
If the value of `result.failed` would be true, no items would have been
filtered from `result.items`.

Change that logic to only keep those items which failed or didn't pass.

### DIFF
--- a/lib/runners/browser_test_runner.js
+++ b/lib/runners/browser_test_runner.js
@@ -83,7 +83,7 @@ BrowserTestRunner.prototype = {
   onTestResult: function(result) {
     var errItems = (result.items || [])
       .filter(function(item) {
-        return !item.passed || result.failed;
+        return !item.passed || item.failed;
       });
 
     this.reporter.report(this.browser, {


johanneswuerbach:
Mind adding a test for this, so we don't regress again? Best place would be https://github.com/testem/testem/blob/master/tests/runners/browser_test_runner_tests.js#L62 I guess.

avdv:
Thinking about it (while writing the test), the logic is not really clear to me.

Because even with the fix, something like this would be allowed:

```json
{
   passed: true,
   failed: true
}
```

Is this what you were after when writing this code? Such that even when passed is true, it would be overruled when `failed` is set to true?

But, what if the passed property is missing:
```json
{
   failed: false
}
```

Since `!result.passed` is true, it doesn't matter what the value of `failed` is.

How about:
```js
return item.passed === false || item.failed;
```

What do you think?

johanneswuerbach:
Unsure, those values are mostly generated by testems adapters, which in most cases forward the test framework values:
https://github.com/testem/testem/blob/master/public/testem/qunit_adapter.js#L88
https://api.qunitjs.com/QUnit.testDone/

I think it would be better to just revert this change and fix the previously broken tests differently.

johanneswuerbach:
Thanks for your PR, I added small test case to your fix :+1: :bow:

avdv:
> Thanks for your PR, I added small test case to your fix

OK, cool. Thanks! :smiley:

PR #773, opened 2016-02-01
add configuration for cdata

jreading:
This PR adds the ability to configure whether or not to include the cdata stack tracs that might not be parsed by some readers.


johanneswuerbach:
Hey, could you add a test for this? Current reporter tests are located here: https://github.com/testem/testem/blob/master/tests/ci/reporter_tests.js

jreading:
Yes, and had I done that I would have seen that false || true equals true. I'm an idiot. *shame*

PR #778, opened 2016-02-03
Wrap Google Chrome with dbus-launch

davidhao3300:
As discussed here: https://github.com/testem/testem/issues/772
I'm not sure how to test for any regressions given that tests appear to be run through mocha and not through any browsers.


johanneswuerbach:
I still don't think this should be part of testem and instead `dbus-launch --exit-with-session testem` should be used. Wouldn't this work?

The reason I tend more to :-1: is that can we be sure that every Linux having `dbus-launch` can wrap chrome like this? What if people using testem in a existing GUI session?

davidhao3300:
Yeah, I'm hoping someone with a Linux browser/anyone who's had problems could speak up.

davidhao3300:
The solution you suggested should work for us, so this PR isn't really critical

davidhao3300:
Maybe better would be to just add something to the docs regarding this issue?

johanneswuerbach:
Yep, I would prefer this. For the current change I just am not able to predict whether this might break anybody or not :cry: .

johanneswuerbach:
Any update on this?

davidhao3300:
I'll add the docs some time next week, have final exams lol

johanneswuerbach:
No worries and good luck with your finals! :-)

manish-northout:
still chromium and chrome are not working through.could you please have a look guys.because locally when i run the test-case its working in Ubuntu 16.04.

johanneswuerbach:
Could you create a new issue with also including additional details about your config and environment? Without this, this is hardly actionable and likely forgotten

PR #792, opened 2016-02-17
Use postMessage for iframe-parent communication

bantic:
The previous code accessed methods and data on `window.parent` directly,
but this is not safe. For instance, setting the `document.domain` within
a test will cause the browser to error when the iframe later attemps to
access a property on `window.parent`. Chrome throws a `SecurityError` that looks like this:

```
Uncaught SecurityError: Blocked a frame with origin "http://localhost:7357" from accessing a frame with origin "http://localhost:7357". The frame being accessed set "document.domain" to "localhost", but the frame requesting access did not. Both must set "document.domain" to the same value to allow access.
```

Phantom has a similar error message:

```
SecurityError: DOM Exception 18: An attempt was made to break through the security policy of the user agent. at http://domain.dev:7357/hello.js, line 2
```

This can be replicated by adding the line `document.domain = document.domain` to the code under test (at least in Chrome, even a redundant change like this will cause it to block attempts to access properties on `window.parent` afterwards).

This changes the code to use `window.postMessage` instead of accessing
`parent` directly. Because message-passing is asynchronous, a fair
amount of code is changing here to better handle that asynchrony. Crucially, message-passing works even if one of the window/iframe changes its `document.domain`.

The iframe will post the following messages to its parent window:
  * `get-id`: The parent will send the id back via another postMessage
  * `reload`: Called when restarting tests, reloads the parent window
  * `iframe-ready`: When the iframe is ready to process messages from
    the parent. Before it receives this event, the parent enqueues
    messages. On receiving 'iframe-ready' it will push all the enqueued
    messages to the iframe (which sends them on via socket.io to the
    api). After receiving this event, messages are no longer enqueued by
    the parent window, they are immediately sent to the iframe.
  * `no-connection-required`: If the id is -1, no socket connection is
    required and the parent need not send the iframe messages intended
    to be relayed over the socket

The iframe listens for these types of messages:
  * `emit-message`: When the parent window is passing it a message to
    relay to the api via socket.io
  * `get-id`: When it receives the id from the parent

To be safe for browsers that do no support postMessage with non-strings as
messages, we use JSON.stringify and JSON.parse on either side of the
postMessage send/receive. Since the messages from the parent to iframe
may contain arbitrarily complex objects (due to `console.log` statements
in test code), we use `decycle` to break cyclical references before
sending the message. This is not necessary (and thus not used) when
posting from the iframe to the parent, because those types of messages and payloads
are not arbitrary (they all originate from the iframe's code).

I tested these changes in several browsers and ran the testem suite in a few different scenarios. I'm not sure if there is a good way to write a test for this specific issue into testem's suite, but if there is I'd be happy to do so.

cc @rwjblue

### DIFF
--- a/public/testem/testem_client.js
+++ b/public/testem/testem_client.js
@@ -12,6 +12,8 @@ It also restarts the tests by refreshing the page when instructed by the server 
 /* globals qunitAdapter, busterAdapter, decycle */
 /* globals Testem */
 /* exported Testem */
+
+var iframe;
 (function appendTestemIframeOnLoad() {
   var iframeAppended = false;
 
@@ -20,7 +22,7 @@ It also restarts the tests by refreshing the page when instructed by the server 
       return;
     }
     iframeAppended = true;
-    var iframe = document.createElement('iframe');
+    iframe = document.createElement('iframe');
     iframe.style.border = 'none';
     iframe.style.position = 'fixed';
     iframe.style.right = '5px';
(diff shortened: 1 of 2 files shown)


bantic:
update: force-pushed a change fixing a jshint error

rwjblue:
@bantic - Thanks for tackling this! This looks great to me...

@johanneswuerbach - These are the changes that we discussed a while back. What do you think?

johanneswuerbach:
This looks just great and I'm happy to merge it as it is when CI says :+1: 

> I tested these changes in several browsers and ran the testem suite in a few different scenarios. I'm not sure if there is a good way to write a test for this specific issue into testem's suite, but if there is I'd be happy to do so.

Currently there isn't :disappointed: Testem CI only runs some E2E saucelabs tests verifying that the scripts work in IE8+ and others: https://github.com/testem/testem/tree/master/examples/saucelabs

I'm definitely interested in building some unit / integration test coverage for those scripts, but couldn't find the time yet.

rwjblue:
The failures in appveyor are hard to parse, it looks like the tests passed (or at least most of them) and we failed on some http-proxy npm config stuff?

bantic:
I'm not sure how to read that either. The `mocha --timeout 5000 tests/*_tests.js tests/**/*_tests.js -R spec` command seems to have run correctly, but something (in the `npm install` process) got caught up in the `node_example` example in the integration test. FWIW, even though I see tests in the `examples/node_example/` directory, when I run `npm run integration` locally it displays 0s (for tests/pass/skip/fail) in the output.

Is it possible to rerun those tests in appveyor with a clean cache?

bantic:
The sporadic "not ok" messages that show up in the mocha test run do appear to be ok — that same output appears when running on master. I think those are messages that are the result of the test itself running a deliberately-failing test suite (to count the number of reported failures/passes).

johanneswuerbach:
No worries, the integration tests are doing several npm installs and are a bit flaky. Retriggered the build, but as one of the jobs was green I'm fine with merging when the next run fails too.

johanneswuerbach:
Thanks again, released as `v1.4.0`

PR #794, opened 2016-02-18
remove name attribute from failure node

### LINKED ISSUE #791: Xunit report should not include name attribute on failure nodes
I'm still having trouble getting Jenkins to parse the xunit report generated by testem. 

I've found [this validator](http://www.freeformatter.com/xml-validator-xsd.html) that fails on the name attribute on the failure element. This is by using [this XSD] (https://svn.jenkins-ci.org/trunk/hudson/dtkit/dtkit-format/dtkit-junit-model/src/main/resources/com/thalesgroup/dtkit/junit/model/xsd/junit-4.xsd) and my report from testem. I can't find any xunit reports that have that attribute.

The name attribute is on the the parent `testcase` element and seems both invalid and redundant. Are you ok with a PR to remove this?

jreading:
Fixes #791

### DIFF
--- a/lib/reporters/xunit_reporter.js
+++ b/lib/reporters/xunit_reporter.js
@@ -72,7 +72,6 @@ XUnitReporter.prototype = {
     var error = result.error;
     if (error) {
       var failureNode = document.createElement('failure');
-      failureNode.setAttribute('name', result.name);
       failureNode.setAttribute('message', error.message);
       if (error.stack && !this.excludeStackTraces) {
         var cdata = document.createCDATASection(error.stack);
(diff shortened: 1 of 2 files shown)


johanneswuerbach:
Could you update the tests? https://github.com/testem/testem/blob/master/tests/ci/reporter_tests.js#L198

jreading:
Done. Modified the existing tests to include the new node for both cases of with and without stack traces. Let me know if you think this is too brittle (testing for absence of `name` seemed wrong).

johanneswuerbach:
Nop, LGTM. Will merge when CI says :+1:

Issue #798, opened 2016-02-23
1.4.0: `iframe.contentWindow` undefined is not object errors


machty:
Upgrading to testem 1.4.0 seems to break things on ember-cli: https://travis-ci.org/machty/ember-concurrency/jobs/111267517

```
not ok 1 PhantomJS 2.1 - Global error: TypeError: undefined is not an object (evaluating 'iframe.contentWindow') at http://localhost:7357/testem.js, line 819
    ---
        Log: |
            { type: 'error',
              text: 'TypeError: undefined is not an object (evaluating \'iframe.contentWindow\') at http://localhost:7357/testem.js, line 819\n' }
```

Downgrading to 1.3.0 fixes this issue.

/cc @rwjblue and @abuiles

johanneswuerbach:
Looks like a race condition when the iframe is not embedded fast enough.

rwjblue:
@johanneswuerbach - Agreed.  Unless someone else tackles this first, I'll try to dig into it tonight...

brancusi:
Getting this as well on any change after initial startup.

bantic:
I will tackle this tonight if @rwjblue doesn't beat me to it. Should be a quick fix.

johanneswuerbach:
Is any of you using `postMessage` in their own app? This is a quick fix, which seems to solve the issue for ember-concurrency: https://github.com/testem/testem/pull/799

bantic:
Looks like ember-concurrency uses [`postMessage` via a dependency on Rx.js](https://github.com/machty/ember-concurrency/blob/master/vendor/dummy-deps/rx.js#L1297-L1307). I've confirmed that's where the too-early messages that trigger the `iframe.contentWindow` problem come from.

machty:
Interesting, I think that's where @abuiles also saw these issues coming from.

johanneswuerbach:
Published as `v1.4.1`

PR #799, opened 2016-02-23
Prevent using iframe before append

johanneswuerbach:
Maybe fixes: https://github.com/testem/testem/issues/798

### DIFF
--- a/public/testem/testem_client.js
+++ b/public/testem/testem_client.js
@@ -185,7 +185,7 @@ function deserializeMessage(message) {
 
 addListener(window, 'message', receiveMessage);
 function receiveMessage(event) {
-  if (event.source !== iframe.contentWindow) {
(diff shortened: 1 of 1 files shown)


bantic:
@johanneswuerbach I think that's a good quick fix. 
The `message` listener on `window` in `testem_client.js` is added before the iframe is appended, as you pointed out in #798. I think those errors are caused by something else posting messages that the listener sees before it has populated its `iframe` variable, i.e. it is receiving messages from something other than the iframe, and it is receiving them before it has appended the iframe.

bantic:
I will take a look later at changing the code so that the `message` listener on window is only attached after the iframe is appended

johanneswuerbach:
Thanks, its quite late here and I don't have time to iterate on it until tomorrow

Issue #732, opened 2016-01-11
Multiple, parallel instances of same launcher?


trentmwillis:
I know it's possible to run parallel instances of _different_ launchers, but is there any way currently to run parallel instances of the _same_ runner? As a bonus, can you run multiple instances with different URLs? I attempted this through various config options but it doesn't seem supported currently.

Specifically, I'm looking to parallelize tests during CI by batching tests across multiple instances of Phantom. Something along the lines of running `/tests?batch=1` on one instance and `/tests?batch=2` on another instance.

Vanuan:
Oh, so parallel option is only for different launchers?
Yeah, sounds obvious, since there's only one html file.

trentmwillis:
@johanneswuerbach just noticed you added a "feature" label to this. I'll see if I can implement something, but I want to discuss the API before beginning work.

I think this should be broken into two features:

1. run parallel instances of the same runner, and
2. run instances with multiple test pages

I tend to think for (1) that this should just be the default if a value greater than `1` is specified for the `parallel` option but only 1 launcher is defined. Unless we want multiple launchers all with multiple instances, in which case I think we should introduce a new config option. I tend to think the second might be the more correct solution, but I'm unsure what that configuration API would look like.

For (2), maybe we define a new config option, `parallel_test_pages` or repurpose the existing `test_page` option such that an array of test page strings get mapped to the launcher instances. Would need a behavior for if the number of test pages doesn't match the number of instances (I tend to think we should opt to using the number of test pages in those cases).

johanneswuerbach:
I really like the array of test pages idea :+1:

The `parallel` setting currently allows you to control your launchers like the following:
* `-1`: All launchers are run in parallel
* `x < launchers`: Launcher runs are pooled and will never exceed `x`, this is important for platforms like saucelabs were you are only allowed to run a limited number of jobs in parallel.
* `x > launchers`: Just runs everything in parallel

I think cleanest way would be to implement running an array of test pages with your defined launchers respecting the defined parallel setting. WDYT?

trentmwillis:
Just to make sure I'm understanding, this means we would essentially use the array of test pages to determine whether multiple instances of a single launcher are run, correct?

So if I had:

```
{
  "test_page": [
    "/tests?batch=1",
    "/tests?batch=2"
  ],
  "launch_in_ci": [
    "PhantomJS"
  ],
  "parallel": -1
}
```

[...]

PR #800, opened 2016-02-23
[FEATURE] Support mutliple test pages in parallel

trentmwillis:
As discussed in #732.

Currently, a work-in-progress, figured I'd put it up for any early advice as I don't have much experience in this code base. It seems to work but sometimes results in unexpected exiting by some of the test duplicate test runners. Wondering if this due to duplicate connections.


examples/multiple_test_pages/tests.js
```diff
@@ -0,0 +1,11 @@
+for (var i = 0; i < 200; i++) {
+  if (window.location.search === '?batch=1') {
```
Vanuan: So it can't split it into batches automatically?

lib/server/index.js
```diff
@@ -223,6 +227,11 @@ Server.prototype = {
     var config = this.config;
     var test_page = config.get('test_page');
 
+    // If there is an array of test pages, we cycle through them as the default
+    if (Array.isArray(test_page)) {
+      test_page = test_page[this.testPageCount % test_page.length];
+    }
```
johanneswuerbach: I would forward always to the first page in this case, as randomly sending launchers might hide the broken behaviour that not every launcher is really testing every batch.

examples/multiple_test_pages/tests.js
trentmwillis: No, that would require testem to have intimate knowledge of how your code is built and run once it is loaded into the browser. This feature is more concerned with giving users the ability to use more than one test page, what they do with those test pages is up to them.

lib/server/index.js
trentmwillis: @johanneswuerbach not quite sure I understand what you mean. The idea here is that each browser launcher should visit each page once. So the first instance of a browser will visit `test_page[0]`, the second will visit `test_page[1]`, and so forth.

I'm not super happy with this solution, but I am unsure of a better way to tell the server what test page should be loaded for a given browser instance.

johanneswuerbach: This approach breaks when you would have  2 browsers, 2 pages and parallel set to 4 as you can't depend on the order of browsers connecting and may end up browser A testing page 1 twice and browser B page 2.

I see two solutions:
1) directly set the test page as launch url per runner
2) store in a map the unique id and the targeted test page at creation time and forward based on that

trentmwillis: Okay, I see now. Good point. I'll attempt method 2, as I expect it will be easier than making sure runner's launch the correct page (since they do not currently navigate directly to a test page).

examples/multiple_test_pages/tests.js
Vanuan: How would it influence "-P" flag? Would user still be able to specify number of parallel runs dynamically? Or is it completely unrelated?

lib/server/index.js
trentmwillis: So I attempted number 2, but am still running into failures. Could it be due to the fact that runners are sharing an instance of a launcher?

It's a bit hard for me to grok all that happens, but it seems like maybe using the same launcher more than once can clobber the socket communications and wind up causing failures. So, rather than creating multiple instances of runners, it looks like maybe I need to create multiple launcher instances where each gets it's own runner, does that sound logical?

johanneswuerbach: Sounds right :+1:

examples/multiple_test_pages/tests.js
stefanpenner: Going forward, we could likely define a convention (batch=1&batch-size=10) or something, then push the popular runners to support this.

lib/server/index.js
stefanpenner: it seems like we would actually want a queue of pending segments to run, and available test runners pop off that queue. This would decouple the number of segments from the actual parallelism, as i can imagine wanted to be able to adjust the parallelism based on several factors (type of machine, current debug task etc.)

Vanuan: @stefanpenner this looks to be in scope of a selenium grid.

trentmwillis:
@stefanpenner I'm not entirely sure I understand what you're getting at.

> a queue of pending segments to run, and available test runners pop off that queue.

It sounds like this is proposing a single, shared queue amongst all test runners, which causes problems if you're expecting every runner (e.g., Phantom and Chrome) to run all pages within the queue. Could you perhaps give an example of what you're proposing?

lib/server/index.js
trentmwillis: @stefanpenner I'm not entirely sure I understand what you're getting at.

> a queue of pending segments to run, and available test runners pop off that queue.

It sounds like this is proposing a single, shared queue amongst all test runners, which causes problems if you're expecting every runner (e.g., Phantom and Chrome) to run all pages within the queue. Could you perhaps give an example of what you're proposing?

examples/multiple_test_pages/tests.js
johanneswuerbach: @Vanuan yes, that should still work and you will never have more then `-P` parallel running launchers.

trentmwillis:
@johanneswuerbach so I got most of this working. It seems like the last issue is that broswers such as Chrome and Safari don't like it when you try to launch more than one instance with the same `user-data-dir`. I can fix this by giving each data directory an id to match the launcher.

However, I want to see if you had an idea for what that API should look like. Primarily since I think it might be able to address #808 as well. I feel like it might mean that the `args` for each browser becomes a function by default.

trentmwillis:
I went ahead and did a simple update to get the `user-data-dir` issue resolved. I think this is ready to go. Tested it extensively on my box (OSX) and didn't notice any issues.

It is worth noting that a couple APIs changed slightly: calling `config.get('test_page')` now returns an array and the `args` option for browser launchers now gets passed the `url` in case it is needed. Not sure if either of those are considered public, but figured I'd highlight it in case there are any concerns about back-compat.

johanneswuerbach:
Sorry for the missed reply, this looks great :+1: Could you also add a E2E integration test here lunching two phantomjs instances and verifying that the reporter has the correct results https://github.com/testem/testem/blob/master/tests/ci/ci_tests.js?

trentmwillis:
Added a new E2E test. Let me know if there's anything else needed, otherwise I'll squash the commits down so we can merge.

johanneswuerbach:
:+1: lgtm

trentmwillis:
Squashed and added a small bit to the `README` for documentation.

johanneswuerbach:
Great, will merge after CI says :+1:

johanneswuerbach:
Released in testem `v1.5.0` :-)

PR #802, opened 2016-02-24
In README, clarifying about '/testem.js'

prad178:
Clarifying that "/testem.js" to be included in custom test page is not the same as "testem.js" defined for nodejs


prad178:
Please make it clear that "/testem.js" is generated by testem on the fly and is not same as the "testem.js" used to run tests.

johanneswuerbach:
Thanks :+1:

PR #807, opened 2016-02-26
cancel existing timeouts

stefanpenner:
* this prevents us from leaking existing timeouts
* this prevents leaked timeouts from spuriously triggering irrelevant timeout exceptions after restarts

- [ ] get feedback
- [x] add tests


johanneswuerbach:
Thanks :clap:

PR #810, opened 2016-02-29
Fixes process runners in dev mode

johanneswuerbach:
Previously the tap and regular process runner weren’t sending the
correct results in dev mode and also not signaling when started or
ended a run

Fixes https://github.com/testem/testem/issues/809


alexindigo:
Now it looks better in the output. :) Thanks.

gavruk:
This works now. But why `spec` reporter shows only 1 test instead of 2?

johanneswuerbach:
As said in the other issue, if you want testem to actually parse mochas tap output, then your config should look like this:

```
"launchers": {
  "Node": {
    "command": "npm run mocha -- --reporter tap tests.js",
    "protocol": "tap"
  }
}
```

The node example is a showcase for spawning any kind of process and returning passed / failed only based on the exit code (so no details / only one test, the exit code)

Maybe we shouldn't use mocha here as that seems to be a bit confusing.

Issue #811, opened 2016-03-01
maxDepth in public/testem/decycle.js


choclaudio:
the payload depth has increased by one but the maxDepth in decycle.js has not reflected this change.
This has the implication that message/stack in errors fall outside of the maxDepth.

Bumbing up maxDepth to 5 (from 4) in decycle.js  solves the issue.

bantic:
Probably related to https://github.com/testem/testem/pull/792.
Can you explain how to view/confirm this error so that I can confirm the fix?

choclaudio:
Hi @bantic, fast response, cool :+1: 

In my case I can create an empty Ember(2.3+) project (uses testem) and just run the tests with any reporter or even just with the --server flag.

so on a vanilla Ember project running
ember-test
ember test --server (look the console output not the web-output)
ember test --reporter teamcity

used to produce the error message and stack trace, but it no longer does.

By bumping up maxDepth in decycle by one the issue is solved, but I do not know if this is the correct place to fix the bug.

choclaudio:
@bantic it might be good if I clarify that I am breaking jshint rules on purpuse to produce a failed test.

PR #813, opened 2016-03-01
fix for issue #811

choclaudio:
bumped up maxDepth from 4 to 5 to solve issue #811


choclaudio:
Please inform me if my PR violates any rules I need to be aware of.

johanneswuerbach:
In general it would be awesome to have a regression test for this, but as its kind of a blocker I'm merging this now without.

//cc @bantic

choclaudio:
@johanneswuerbach Agreed, would be nice to catch any regression on this issue. I am pretty new to the open-source world and also test-frameworks in JavaScript(Java/C# background), but I might muster up a regresion-test for it.

PR #823, opened 2016-03-07
update to support asynchronous callbacks from the browser

### LINKED ISSUE #360: Hard to listen to all-test-results event
The code of testem client is

```js
window.Testem = {
  emit: function(evt){
    var args = Array.prototype.slice.apply(arguments)
    socket.emit.apply(socket, circularSafe(args))
    if (this.evtHandlers && this.evtHandlers[evt]){
      var handlers = this.evtHandlers[evt]
      var args = Array.prototype.slice.call(arguments, 1)
      for (var i = 0; i < handlers.length; i++){
        var handler = handlers[i]
        handler.apply(this, args)
      }
    }
  },
```

Which means, first the socket then the callbacks. In case of `all-test-results`, the socket message tells `testem ci` to kill the browser, so depending on how quick this happen, the handler callback might not be called.

It would be good to move the socket.emit at the end of the method

jreading:
First pass at having asynchronous callbacks from the browser. 
Fixes #360

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,6 +409,33 @@ If you need to run a command after your tests have completed (such as removing c
 
 If you would prefer simply to clean up when Testem exits, you can use the `on_exit` option.
 
+Running browser code after tests complete
+-------------
+It is possible to send coverage reports or run other javascript in the browser by using the `afterTests` method.
+
+```javascript
+Testem.afterTests(
+  //Asynchronously
+  function(config, data, callback) {
+    var coverage = window.__coverage__;
+    var postBody = JSON.stringify(coverage);
+    if (postBody) {
+        var xhr = new XMLHttpRequest();
+        xhr.onreadystatechange = function() {
+            if (xhr.readyState === 4) {
+                callback();
+            }
+        };
+        xhr.open('POST', 'http://localhost:7358/', true);
+        xhr.send(postBody);
+    }
+});
+
+//Synchronously
+Testem.afterTests(doStuff);
+```
+
+
 Custom Routes
 -------------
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -409,6 +409,34 @@ If you need to run a command after your tests have completed (such as removing c
...
 If you would prefer simply to clean up when Testem exits, you can use the `on_exit` option.
 
+Running browser code after tests complete
+-------------
+It is possible to send coverage reports or run other javascript in the browser by using the `afterTests` method. For asynchronous functions, just run a callback function as the first argument.
+
+```javascript
+Testem.afterTests(
+  //Asynchronously
+  function(cb) {
```
johanneswuerbach: I would prefer if we use the same API as you can already use in the config `function(config, data, callback)` to allow passing more additional details in the future.

https://github.com/testem/testem/blob/784e210f0619fcebd32f5ac51062396597a38e5f/examples/dynamic_config/testem.js#L16

I would also drop the `{"async": true}` and require the callback by default.

jreading:
This should be ready now.

johanneswuerbach:
Nearly, could you also update this example https://github.com/testem/testem/tree/master/examples/coverage_istanbul to use your technic? Examples are continuously tested, so this should improve the test coverage here a bit.

Otherwise, LGTM :+1:

johanneswuerbach:
Thanks :+1: could you squash your commits then I can merge!

johanneswuerbach:
Looks like the test failures are related, could you have a look?

jreading:
tap runner fixed.

Issue #821, opened 2016-03-04
Can no longer set custom reporter in testem.js


dbushong:
As of https://github.com/testem/testem/commit/1ea331eaa192b737722ec8154d8839f6e56d2d03 you can no longer configure a custom `reporter` in `testem.js`:

The priority of `fileOptions` is lower than that of `progOptions`, and `progOptions` has a default for `-R` (`tap`), thus no matter what you put in `testem.js` for `reporter`, it never gets used.

johanneswuerbach:
What is defaulting `progOptions` to `tap`? Are you using testem standalone or with `ember-cli`?

dbushong:
https://github.com/testem/testem/blob/master/testem.js#L36 standalone

gwenaelp:
I have this issue too. If you unset the default option in https://github.com/testem/testem/blob/master/testem.js#L36, it works like a charm.

Any plans to handle this upstream?

johanneswuerbach:
Sure, PRs welcome. I'll have a look as soon as I have time.

DerivcoHK:
It is because the "getConfigProperty" function of "Config" will return the progOptions first rather than that of fileOptions. 

```
Config.prototype.getConfigProperty = function (key) {
  ....
  if (key in this.progOptions && typeof this.progOptions[key] !== 'undefined') {
    return this.progOptions[key];
  }

  if (key in this.fileOptions&& typeof this.fileOptions[key] !== 'undefined') {
    return this.fileOptions[key];
  }
  .....
}
```

So when we unset the default option in https://github.com/testem/testem/blob/master/testem.js#L36, it works.

johanneswuerbach:
Fixed by https://github.com/testem/testem/pull/860

PR #825, opened 2016-03-08
issue 821 fix proposal

gwenaelp:
make reporter default option stop overriding reporter option in configuration files, to allow defining custom reporters again (issue #821)

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -33,7 +33,7 @@ function App(config, finalizer) {
 
   this.reportFileName = this.config.get('report_file');
   this.reportFileStream = this.initReportFileStream(this.reportFileName);
-  this.reporter = this.initReporter(this.config.get('reporter'), this.reportFileStream);
+  this.reporter = this.initReporter((this.config.get('reporter') || 'tap'), this.reportFileStream);
   if (!this.reporter) {
     this.cleanExit(new Error(
       'Test reporter `' + this.config.get('reporter') + '` not found.'
--- a/testem.js
+++ b/testem.js
@@ -33,7 +33,7 @@ program
   .option('-T, --timeout [sec]', 'timeout a browser after [sec] seconds', null)
   .option('-P, --parallel [num]', 'number of browsers to run in parallel, defaults to 1', Number)
   .option('-b, --bail_on_uncaught_error', 'Bail on any uncaught errors')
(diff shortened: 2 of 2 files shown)


lib/app.js
```diff
@@ -33,7 +33,7 @@ function App(config, finalizer) {
 
   this.reportFileName = this.config.get('report_file');
   this.reportFileStream = this.initReportFileStream(this.reportFileName);
-  this.reporter = this.initReporter(this.config.get('reporter'), this.reportFileStream);
+  this.reporter = this.initReporter((this.config.get('reporter') || 'tap'), this.reportFileStream);
```
johanneswuerbach: This should be required as `tap` is still the default https://github.com/testem/testem/blob/master/lib/config.js#L160.

johanneswuerbach:
Could you add a test so this won't break again? :-)

DerivcoHK:
Can you please merge this fix as soon as possible ?

johanneswuerbach:
Not without tests to prevent further regressions. My time is a bit limited atm, but PRs are welcome.

DerivcoHK:
**config.js**
```
Config.prototype.defaults = {
  ....
  reporter: 'tap',
   ....
};
```

```
Config.prototype.getConfigProperty = function (key) {
  if (this.config && key in this.config) {
    return this.config[key];
  }
  if (key in this.progOptions && typeof this.progOptions[key] !== 'undefined') {
    return this.progOptions[key];
  }
  if (key in this.fileOptions && typeof this.fileOptions[key] !== 'undefined') {
    return this.fileOptions[key];
  }
  if (key in this.defaults) {
     ....
      return defaultVal;
  }
}
```

**testem.js**
`  .option('-R, --reporter [reporter]', 'Test reporter to use [tap|dot|xunit]', 'tap')
`

Now the priority of progOptions is higher than that of fileOptions, and we have default progOptions for reporter 'tap' so the "reporter" option in config cannot be used anymore. 

It is safe to remove the default value 'tap' in the -R option, because it will get the default value when there are no progOptions and fileOptions.

How can I add a test to show it is safe and so that you can merge it?

johanneswuerbach:
Make sense, could you send a PR just removing the `tap`?

johanneswuerbach:
Fixed by https://github.com/testem/testem/pull/860

PR #849, opened 2016-04-07
Change teamcity reporter to remove reference to mocha

somewhatabstract:
I was using this reporter with jasmine and wondered why it referenced mocha. Thought having it read "testem" would be more appropriate.

### DIFF
--- a/lib/reporters/teamcity_reporter.js
+++ b/lib/reporters/teamcity_reporter.js
@@ -26,7 +26,7 @@ TeamcityReporter.prototype = {
     }
     this.endTime = new Date();
     this.out.write('\n\n');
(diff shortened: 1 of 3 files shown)


johanneswuerbach:
Good catch, the test failure seems related. Could you a look?

somewhatabstract:
Certainly. On it.

somewhatabstract:
I think that should cover it (no pun intended)

somewhatabstract:
That one build failing doesn't appear to be down to this change per se. Any ideas?

somewhatabstract:
Nevermind that last comment. Looks like it's all working now.

PR #860, opened 2016-04-19
Remove the default progoption of reporter

DerivcoHK:
Remove the default progOption of reporter, to make the custom reporter work.

It will not affect the default value of reporter, cause the Config will get the default value "tap" when there are no progOption and fileOption of reporter.

### DIFF
--- a/testem.js
+++ b/testem.js
@@ -33,7 +33,7 @@ program
   .option('-T, --timeout [sec]', 'timeout a browser after [sec] seconds', null)
   .option('-P, --parallel [num]', 'number of browsers to run in parallel, defaults to 1', Number)
   .option('-b, --bail_on_uncaught_error', 'Bail on any uncaught errors')
-  .option('-R, --reporter [reporter]', 'Test reporter to use [tap|dot|xunit|teamcity]', 'tap')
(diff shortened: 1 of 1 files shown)


DerivcoHK:
**config.js**
```
Config.prototype.defaults = {
  ....
  reporter: 'tap',
   ....
};
```

```
Config.prototype.getConfigProperty = function (key) {
  if (this.config && key in this.config) {
    return this.config[key];
  }
  if (key in this.progOptions && typeof this.progOptions[key] !== 'undefined') {
    return this.progOptions[key];
  }
  if (key in this.fileOptions && typeof this.fileOptions[key] !== 'undefined') {
    return this.fileOptions[key];
  }
  if (key in this.defaults) {
     ....
      return defaultVal;
  }
}
```

**testem.js**
`  .option('-R, --reporter [reporter]', 'Test reporter to use [tap|dot|xunit]', 'tap')
`

Now the priority of progOptions is higher than that of fileOptions, and we have default progOptions for reporter 'tap' so the "reporter" option in config cannot be used anymore. 

It is safe to remove the default value 'tap' in the -R option, because it will get the default value when there are no progOptions and fileOptions.

PR #866, opened 2016-05-07
Passing metadata from an adapter to a reporter.

### LINKED ISSUE #863: Passing Metadata from Test to Reporter
I would like to collect some information from my running tests and surface them to my custom reporter. What I'm looking for is something similar to logs, except something that allows objects through, possibly with a category or type (as a string) to allow more structured filtering.

The flow I am thinking of is: adapter emits a new type of event, test runner collects it and passes it along to the reporter. Open to other ideas as well.

Willing to work on a PR if there ends up being agreement that this is the right direction and approach.

tjni:
Resolves #863.

While this works, I need to come up with an acceptable way to relax or customize the max depth used in decycle, and haven't come up with a clean way to do this yet. Suggestions welcome!


tjni:
The second commit is a possible way to handle the max depth configuration.

tjni:
The max depth configuration I tried out doesn't work when messages to the iframe are being queued up to be emitted later. I've grown to like the granularity of configuring max depth per adapter, so I will spend some time trying refactor and make it work. Stay tuned.

johanneswuerbach:
Thanks, looks already great 👍

tjni:
It works now from my manual testing. I'd like to write some tests, but I'm not sure where the best place is for testing the code in testem_client.js.

tjni:
I was able to return to this today and made an attempt to add some tests to the changes that allows customizing the decycle depth. Going to work on merging cleanly and getting the build to pass.

tjni:
I think this is at last ready to be reviewed.

The second commit required moving some sections of code around in order to satisfy ESLint, hence why a couple of lines have been shuffled around.

johanneswuerbach:
👏 Great PR. If you are ready, I would be fine with merging as it is.

tjni:
Yes, that would be awesome. Thank you!

PR #891, opened 2016-06-10
Update README.

fotinakis:
I found out the hard way that you must accept all three arguments to any afterTests method and call the `callback()` method, otherwise tests will hang. It would be nice if this behavior didn't exist, but a comment is at least called for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,9 +431,11 @@ Testem.afterTests(
     }
 });
 
-//Synchronously
+// Synchronously.
 Testem.afterTests(doStuff);
 ```
+
+NOTE: You must call the `callback()` argument or your tests will hang.
 
 
 Custom Routes


fotinakis:
And apparently this doesn't actually go far enough, I can't even get `afterTests` to work with a top-level function because it throws `Cannot read property 'afterTestsQueue' of undefined` (referring to `this.afterTestsQueue` in: https://github.com/testem/testem/blob/7a89cabdb20741e7801f8eeb3a0c7e3cbea0b869/public/testem/testem_client.js#L307). :|

I've resorted to using `Testem.on('after-tests-complete', doStuff);` instead of the `afterTests` hook. Feel free to merge or close this PR, but it seems like `afterTests` behavior needs to be audited.

johanneswuerbach:
How are you calling the callback? Looks like the scope is somehow messed up. 

You can find an example how this should work is here https://github.com/testem/testem/blob/master/examples/coverage_istanbul/tests.html#L11, does that work for you?

> It would be nice if this behaviour didn't exist, but a comment is at least called for.

I can't really think of a different way of doing async work, ideas?

fotinakis:
I was doing something like:

```javascript
function finalize(config, data, callback) {
  // ...
  if (callback) {
    callback();
  }
}

export default function(name) {
  if (window.Testem.afterTests) {
    // Testem >= v1.6.0
    window.Testem.afterTests(finalize)
  } else {
    // Testem < v1.6.0.
    window.Testem.on('all-test-results', finalize);
  }
  // ...
}
```

This was wrapped as an async test helper for ember-cli, which I believe was pinned to testem 1.6.0 at the time. I tried a bunch of different binding things that I thought would fix the `Cannot read property 'afterTestsQueue' of undefined` error but ended up just dropping it.

Perhaps this is as simple as switching `this.afterTestsQueue` to `Testem.afterTestsQueue` instead of using them both mixed?

fotinakis:
Since I have a workaround using `Testem.on('after-tests-complete', doStuff);` instead of `afterTests`, going to close this for now. If I encounter this again I'll dig into it more to provide you with better info—but I do think there are `this`-scoping bugs present here: https://github.com/testem/testem/blob/1.8.x/public/testem/testem_client.js#L308

PR #905, opened 2016-06-22
Changed runAfterTests shift to use the same object as the conditional 

travis-jm:
Without this change, you have to do `callback.apply(Testem)` in your `.afterTests` to prevent `this` from being undefined.


johanneswuerbach:
Great 👏 , could you add a test here https://github.com/testem/testem/blob/master/tests/client_tests.js so we don't regress again?

PR #911, opened 2016-06-28
Make test run stopping depend on process cleanup.

### LINKED ISSUE #902: Version 1.9.0 regression
Testem crash when I run it in ```dev``` mode. Here is my testem.js:
```javascript
module.exports = {
  framework: 'mocha+chai',
  src_files: [
    'src/**/*.js',
    'test/setup.js',
    'test/**/*.spec.js'
  ],
  launchers: {
    node: {
      command: 'mocha -R tap',
      protocol: 'tap'
    }
  },
  launch_in_dev: [ 'node' ]
};
```

Here is the log I get when running with ```./node_modules/.bin/testem --debug```:
```
info Test'em starting.. 
info Seeking for config file... 
info Starting dev 
info Starting server 
info node 
info runnerAdded 
info Start triggered test run. 
info Running tests... 
info node onStart
info cmd: mocha -R tap 
info executing: mocha -R tap 
info ok 1 Is this a test yes it is
 
WARN node closed 0
info src/index.js changed (enabled). 
info test/dummy.spec.js changed (enabled). 
info File changed: src/index.js triggered test run. 
info File changed: test/dummy.spec.js triggered test run. 
info Process node already killed. 
info Running tests... 
info Process node already killed. 
```

The ```./node_modules/.bin/testem --debug``` command exits with code ```1```.

This works as expected with 1.8.1. Here is the debug output in 1.8.1:
```
info Test'em starting.. 
info Seeking for config file... 
info Starting dev 
info Starting server 
info node 
info runnerAdded 
info Start triggered test run. 
info Process node already killed. 
info Running tests... 
info node onStart
info cmd: mocha -R tap 
info executing: mocha -R tap 
info ok 1 Is this a test yes it is
 
WARN node closed 0
info src/index.js changed (enabled). 
info File changed: src/index.js triggered test run. 
info Process node already killed. 
info Running tests... 
info test/dummy.spec.js changed (enabled). 
info File changed: test/dummy.spec.js triggered test run. 
info Process node already killed. 
info Running tests... 
info node onStart
info node onStart
info cmd: mocha -R tap 
info executing: mocha -R tap 
info cmd: mocha -R tap 
info executing: mocha -R tap 
info node onEnd
info ok 1 Is this a test yes it is
 
info ok 1 Is this a test yes it is
 
WARN node closed 0
```

It runs the test suite multiple times, but that's the only odd thing. The terminal still displays the UI, waiting for files to change or 'ENTER'/'q'/'p' to be pressed.

If you need any more info or want me to run with some more switches enabled, let me know.

tjni:
Tries to fix #902.

Manual testing using a sample small app suggests that this change fixes the crash. Logically, I think it is correct that the stopping flag is only reset after the launcher processes have a chance to clean up. I removed the node style callbacks from the two methods I touched, since they are not currently needed.

I spent several minutes trying to figure out a clean way to write a test that fails before and passes after, but could not come up with anything that doesn't use timing, which always makes me uncomfortable. I'm open to suggestions on how to write an automated regression test for this; in the meantime, would love to see this merged if the change makes sense.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -111,19 +111,16 @@ App.prototype = {
     });
   },
 
-  stopCurrentRun: function(cb) {
+  stopCurrentRun: function() {
     if (this.stopping || !this.currentRun) {
       return Bluebird.resolve();
     }
     this.stopping = true;
 
     var self = this;
-
-    this.cleanUpProcessLaunchers();
-
-    return this.currentRun.finally(function() {
+    return Bluebird.all([ this.cleanUpProcessLaunchers(), this.currentRun ]).finally(function() {
       self.stopping = false;
-    }).asCallback(cb);
+    });
   },
 
   runTests: function(err, cb) {
@@ -404,13 +401,13 @@ App.prototype = {
     return null;
   },
 
-  cleanUpProcessLaunchers: function(callback) {
+  cleanUpProcessLaunchers: function() {
     var processRunners = this.runners.filter(function(runner) {
       return runner.launcher.isProcess();
     });
     return Bluebird.each(processRunners, function(runner) {
       return runner.stop();
-    }).asCallback(callback);
+    });
   },
 
   cleanUpLaunchers: function(callback) {


johanneswuerbach:
I'm not sure that this is the right fix as `stopCurrentRun` shouldn't resolve before the `currentRun` is finished.

While the code was not directly waiting for `cleanUpProcessLaunchers`, it was indirectly as this map should resolve when launchers are killed and skip all outstanding launchers: https://github.com/testem/testem/blob/master/lib/app.js#L354

Looks a bit like a race condition, were you able to get a stack trace of a crash?

tjni:
I didn't manage to find a stack trace and instead had to fall back on the tried and true strategy of logging things all over the place. I'm still not entirely sure what exactly is causing the crash.

I can rationalize why not waiting for `cleanUpProcessLaunchers` is an issue, however. That method call kicks off some process killing asynchronously, and I don't see code in `singleRun` waiting for this. In fact, it resolves even sooner, as soon as it sees that the intention to stop (via the stopping flag).

Perhaps the proper solution is to wait for both `cleanUpProcessLaunchers` and `currentRun` to resolve before resetting the stopping flag?

johanneswuerbach:
Still not fully understand why this required but fine with merging as it shouldnt hurt.

tjni:
I'm not sure I can give a great answer, but I'll describe in more detail the sequence I think is happening:

1. File watcher triggers a new run by calling `triggerRun`.
2. `triggerRun` calls `stopCurrentRun` and waits for it to resolve.
3. `stopCurrentRun` sets the `stopping` flag to true, calls `cleanUpProcessLaunchers`, adds logic to the `currentRun` promise to reset the `stopping` flag after it resolves, and then returns the current run.
4. Back in `triggerRun`, a new test run is started.

Now, `cleanUpProcessLaunchers` tells the launcher to kill itself by sending a signal, which happens asynchronously. It returns a promise that resolves after the process dies.

If `stopCurrentRun` doesn't wait for the `cleanUpProcessLaunchers` to resolve -- and this is the part where I'm hazy -- the processes being killed might interfere with the new run that is starting.

PR #912, opened 2016-06-29
Chrome Canary uses 'Canary' in the binary name

kbullaughey:


lib/utils/known-browsers.js
```diff
@@ -15,7 +15,7 @@ function chromeWinPaths(homeDir, name) {
 function chromeOSXPaths(name) {
   return [
     process.env.HOME + '/Applications/' + name + '.app/Contents/MacOS/Google\ Chrome',
```
johanneswuerbach: Thanks, could you also fix it here?

kbullaughey: Missed that. Done.

PR #919, opened 2016-07-07
Use forked fireworm

### LINKED ISSUE #785: Zillions of Chrome tabs opened
The bug in fireworm causes testem to detect false file modifications, resulting in a new browser instance launched for each file.

Please update testem as soon as this is merged and a new version of fireworm released:
https://github.com/airportyh/fireworm/pull/10

Vanuan:
Fixes #785


johanneswuerbach:
I would prefer to move to a more supported file-watcher instead, maybe https://github.com/amasad/sane, would you have time to create a PR replacing fireworm with sane?

Vanuan:
Would it be possible to let the user specify the watch command or even decouple them?
Testem's interface is http server, right? And it always runs http server, right? So it would be possible to send the signal to rerun tests externally?

For example, I'd want nodemon. Testem would provide me the binary that I should run to trigger rerun. And I'll use it like this:
```
testem                                   # run testem server
nodemon --exec testem-rerun *.js         # run my custom watcher
```
testem-rerun would just send the command through http. Though it would need the way to specify http port somehow... Probably testem would write http port number to some dedicated location...

In ci mode I don't need watcher, so none will be provided.

What do you think?

johanneswuerbach:
Interesting idea, testem could definitely listed to some signal event https://nodejs.org/api/process.html#process_signal_events so other process could trigger a reload.

As filewatching is already implemented I wouldn't really remove it, just move to a maintained file watcher.

Vanuan:
Sorry didn't have much time. Have to use forked testem for now.

johanneswuerbach:
Replaced by https://github.com/testem/testem/pull/1981

Issue #924, opened 2016-07-18
various timers appear to not be cleaned up


stefanpenner:
For example, it seems like: https://github.com/testem/testem/blob/master/lib/runners/browser_test_runner.js#L204

In some scenarios does not get cleaned up, which results in (when invoking testem in-proccess) artificial delays of up to 10s before the process exits.

using `wtfnode` resulted in:

```
[WTF Node?] open handles:
- File descriptors: (note: stdio always exists)
  - fd 1 (tty) (stdio)
  - fd 2 (tty) (stdio)
- Timers:
  - (10000 ~ 10 s) (anonymous) @ /Users/spenner/src/testem/testem/lib/runners/browser_test_runner.js:204
  - (1000 ~ 1000 ms) (anonymous) @ /Users/spenner/src/testem/testem/lib/runners/browser_test_runner.js:224
```

Anyways, I'll dig in further. Figured I would open an issue to log my findings and related progress.

stefanpenner:
^^ PR that should address the issue, would <3 some 👀  on it.

PR #925, opened 2016-07-18
[Bugfix #924] prevent unnecessary waits on shutdown

stefanpenner:
* don’t schedule “cleanup” tasks if already finished
* store setTimeout cookies, and cancel when finished

----

[Bugfix #924]

### DIFF
--- a/lib/runners/browser_test_runner.js
+++ b/lib/runners/browser_test_runner.js
@@ -13,6 +13,9 @@ function BrowserTestRunner(launcher, reporter, index, singleRun) {
   this.id = this.launcher.id;
   this.singleRun = singleRun;
   this.logs = [];
+
+  this.pendingTimer = undefined;
+  this.onProcessExitTimer = undefined;
 }
 
 BrowserTestRunner.prototype = {
@@ -200,6 +203,9 @@ BrowserTestRunner.prototype = {
   },
   onDisconnect: function() {
     var self = this;
+
+    if (this.finished) { return; }
+
     this.pending = true;
     this.pendingTimer = setTimeout(function() {
       if (self.finished) {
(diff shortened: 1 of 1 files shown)


johanneswuerbach:
Any idea how to test this?

johanneswuerbach:
I think this totally makes sense, but all the boolean flags combinations in this file get more and more confusing. Fine with merging as it is for now.

stefanpenner:
@johanneswuerbach we could use `process._getActiveHandles()` and `process._getActiveRequests()` inspect them before running testem, and after and make sure they  remain the same (e.g. testem doesn't leave anything lingering)...

stefanpenner:
@johanneswuerbach any recommendation which test case I should start with to do ^, or are you interested in writing the test

johanneswuerbach:
Verifying the handles seems like a good approach. I would add it somewhere around here: https://github.com/testem/testem/blob/master/tests/ci/ci_tests.js#L96

> or are you interested in writing the test
Sadly don't really have time 😞

stefanpenner:
@johanneswuerbach thanks for the hint, I'll write the test.

stefanpenner:
@johanneswuerbach as for improving the code quality, we may want to use some better primitives that abstract these types of problems more... Obviously neither of us have time right now, but food for future refactorings.

stefanpenner:
@johanneswuerbach adding the tests is a-bit tricky, due to ambient tasks adding/removing handles, I believe I have a  reasonable solution...

johanneswuerbach:
Change and code looks good, sadly it looks like there is another leaks.

stefanpenner:
@johanneswuerbach I think that test mechanism is too brittle. Without more refactoring, Im not sure we can do much that wont itself be more tech debt.

Unfortunately for now, we much just want to 🚢  this. Maybe we will have time soon to refactor..

johanneswuerbach:
👍

PR #946, opened 2016-08-06
Properly handle flags in query params

trentmwillis:
Currently, if you pass a "flag" as a query parameter (e.g., a key without a value), then the resultant URL contains an unnecessary equal sign. In other words:

```js
var qs = require('querystring');
qs.stringify(qs.parse('foo&bar')) // => 'foo=&bar='
```

This is troublesome for query parsing implementations that I've worked with as the intent is unclear. Additionally, it results in unexpected behavior as a passed in query string may not be properly reflected in the browser.

Unfortunately, it doesn't look like this will be [supported by Node itself](https://github.com/nodejs/node/issues/7330) in the near future as the standard they've adopted does not define a behavior for it. So, I've implemented a simple solution within the Config code and modified the appropriate tests to cover this case.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -187,7 +187,9 @@ Config.prototype.mergeUrlAndQueryParams = function(urlString, queryParamsObj) {
     outputQueryParams[param] = queryParamsObj[param];
   });
   urlObj.query = outputQueryParams;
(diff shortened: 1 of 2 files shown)


johanneswuerbach:
Thanks, lgtm

PR #949, opened 2016-08-06
Rm node 5

johanneswuerbach:
Node 5 is EOL. https://github.com/nodejs/LTS#lts_schedule

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,15 +7,14 @@ node_js:
   - "0.10"
   - "0.12"
   - "4"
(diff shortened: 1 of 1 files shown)


jdalton:
Heads up Node 0.10 hit its EOL recently.

johanneswuerbach:
Yes, I was thinking of keeping it for two more months until 0.12 also hits EOL and then remove both and release testem v2.

WDYT?

jdalton:
Sounds rad!

PR #950, opened 2016-08-06
Use individual lodash function

bmeurant:
Switch to individual lodash function in production code to reduce dependencies.

note that ``lodash.extend`` is an alias for ``lodash.assignIn`` (see [here](https://lodash.com/docs#assignIn))

This will save 554 dep files in production (npm2 mode):

```
# with full lodash
npm install --legacy-bundling --only=production
find node_modules/ -type f | wc -l
2499

# with individual lodash
npm install --legacy-bundling --only=production
find node_modules/ -type f | wc -l
1945
```


bmeurant:
part of the quest ember-cli/ember-cli#6149

johanneswuerbach:
Thanks for doing that, looks like lodash can be ditched nearly all together.

PR #954, opened 2016-08-11
Add dummy error handler for proxies

alltouch:
In our project we have following error:
client aborting request to proxy => proxy raises error => test fails => we unhappy

The only solution i know add error handler. Implemented simple code for it.
@johanneswuerbach, what do you think about it?

### DIFF
--- a/lib/server/index.js
+++ b/lib/server/index.js
@@ -187,6 +187,14 @@ Server.prototype = {
     if (proxies) {
       self.proxy = new httpProxy.createProxyServer({changeOrigin: true});
 
+      self.proxy.on('error', function (err, req, res) {
+        console.error(req);
+        res.writeHead(500, {
+          'Content-Type': 'text/plain'
+        });
+
+        res.end('Something went wrong. And we are reporting a custom error message.');
+      });
       Object.keys(proxies).forEach(function(url) {
         app.all(url + '*', function(req, res, next) {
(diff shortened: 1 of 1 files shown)


lib/server/index.js
```diff
@@ -187,6 +187,13 @@ Server.prototype = {
     if (proxies) {
       self.proxy = new httpProxy.createProxyServer({changeOrigin: true});
 
+      self.proxy.on('error', function (err, req, res) {
+        res.writeHead(500, {
+          'Content-Type': 'text/plain'
+        });
+
+        res.end('Something went wrong. And we are reporting a custom error message.');
```
johanneswuerbach: Feels like we should also log the error itself. Maybe sentry.captureErorror?

akatov: where does testem use sentry?

johanneswuerbach: Sorry, I got confused. Logging the error itself and the stack should be fine for now. As testem shouldn't be used in production the stack can also be part of the response for easier development

alltouch:
@johanneswuerbach
Updated code. Do we need any other changes?

johanneswuerbach:
Sorry for letting this linger so long, would you mind adding a test here https://github.com/testem/testem/blob/master/tests/server_tests.js? Then this is ready to merge.

PR #959, opened 2016-08-22
Launchers fix: replace exec with spawn for command

chrmod:
https://github.com/testem/testem/issues/958

### DIFF
--- a/lib/process-ctl.js
+++ b/lib/process-ctl.js
@@ -6,6 +6,7 @@ var crossSpawn = require('cross-spawn');
 var Bluebird = require('bluebird');
 var util = require('util');
 var EventEmitter = require('events').EventEmitter;
+var spawnargs = require('spawn-args');
 
 var envWithLocalPath = require('./env-with-local-path');
 var fileutils = require('./fileutils');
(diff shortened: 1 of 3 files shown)


lib/process-ctl.js
```diff
@@ -86,9 +86,17 @@ ProcessCtl.prototype.exec = function(cmd, options) {
   log.info('executing: ' + cmd);
 
   var self = this;
-  this.process = childProcess.exec(cmd, _options, function(err, stdout, stderr) {
-    self.stdout = stdout;
-    self.stderr = stderr;
+  var cmdParts = cmd.split(" ");
```
johanneswuerbach: This won't support quoted commands like: `node -e "console.log('with spaces');"`

chrmod:
windows issues seems to come from *spawn-args*. @johanneswuerbach any suggestions how to deal with them?

johanneswuerbach:
Looks like `spawn-args` works fine, great find 👏  I think the test can be fixed by replacing this line https://github.com/testem/testem/blob/master/tests/launcher_tests.js#L74 with

`assert.equal(stdout, 'hello\n');`

Not entirely sure why, but the same test using `spawn` is also using `\n` instead of `\r\n` https://github.com/testem/testem/blob/master/tests/launcher_tests.js#L158

chrmod:
@johanneswuerbach all tests green

johanneswuerbach:
Thanks 👍

PR #965, opened 2016-08-31
fix teamcity reporter - handle undefined errors

calderas:
When there's a global error the reporter breaks the process and exits.
This PR just adds a validation to prevent throwing an error within the reporter when the test error is undefined.

Currently you will get something like this:
```
TypeError: Cannot read property 'message' of undefined
      at Object.TeamcityReporter._display (lib/reporters/teamcity_reporter.js:41:124)
      at Object.TeamcityReporter.report (lib/reporters/teamcity_reporter.js:17:10)
      at Context.<anonymous> (tests/ci/reporter_tests.js:406:16)
npm ERR! Test failed.  See above for more details.
Process exited with code 1
```

The actual result looks something like this and the error is **undefined**:
```
{ passed: false,
  name: 'Global error: ReferenceError: Can\'t find variable: XYZ at http://localhost:7357/668/tests/index.html?hidepassed=&_split=24&_partition=11, line 38\n',
  skipped: undefined,
  runDuration: undefined,
  logs:
   [ { type: 'error',
       text: 'ReferenceError: Can\'t find variable: XYZ at http://localhost:7357/668/tests/index.html?hidepassed=&_split=24&_partition=11, line 38\n' } ],
  error: undefined,
  launcherId: '668',
  failed: 1,
  pending: undefined,
  items: undefined }
```

### DIFF
--- a/lib/reporters/teamcity_reporter.js
+++ b/lib/reporters/teamcity_reporter.js
@@ -38,7 +38,9 @@ TeamcityReporter.prototype = {
     if (result.skipped) {
       this.out.write('##teamcity[testIgnored name=\'' + this._namify(prefix, result) + '\' message=\'pending\']\n');
     } else if (!result.passed) {
-      this.out.write('##teamcity[testFailed name=\'' + this._namify(prefix, result) + '\' message=\'' + escape(result.error.message) + '\' details=\'' + escape(result.error.stack) + '\']\n');
+      var message = (result.error && result.error.message) || '';
+      var stack = (result.error && result.error.stack) || '';
+      this.out.write('##teamcity[testFailed name=\'' + this._namify(prefix, result) + '\' message=\'' + escape(message) + '\' details=\'' + escape(stack) + '\']\n');
     }
     this.out.write('##teamcity[testFinished name=\'' + this._namify(prefix, result) + '\']\n');
 
--- a/tests/ci/reporter_tests.js
+++ b/tests/ci/reporter_tests.js
@@ -402,12 +402,40 @@ describe('test reporters', function() {
         name: 'it skips stuff',
         skipped: true
       });
+
+      reporter.report('phantomjs', {
+        name: 'it handles failures',
+        passed: false,
+        error: {
+          passed: false,
+          message: 'foo',
+          stack: 'bar'
+        }
+      });
+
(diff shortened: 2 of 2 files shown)


lib/reporters/teamcity_reporter.js
```diff
@@ -38,7 +38,9 @@ TeamcityReporter.prototype = {
     if (result.skipped) {
       this.out.write('##teamcity[testIgnored name=\'' + this._namify(prefix, result) + '\' message=\'pending\']\n');
     } else if (!result.passed) {
-      this.out.write('##teamcity[testFailed name=\'' + this._namify(prefix, result) + '\' message=\'' + escape(result.error.message) + '\' details=\'' + escape(result.error.stack) + '\']\n');
+      var message = (result.error && result.error.message) || '';
+      var stack = (result.error && result.error.stack) || '';
+      this.out.write('##teamcity[testFailed name=\'' + this._namify(prefix, result) + '\' message=\'' + escape(message) + '\' details=\'' + escape(stack) + '\']\n');
```
johanneswuerbach: Should `message=` be remove when the key doesn't exists? Don't really know how teamcity will render this.

calderas: It just renders empty message & details

```
##teamcity[testFailed name='PhantomJS 2.1 - Global error: ReferenceError: Can|'t find variable:' message='' details='']
``` 
https://github.com/testem/testem/pull/965/commits/2240b27090306e06b8b721f87c2d6879f86f80e3#diff-b0131a2a1e9b5b26022dfe678ebd3e6dR437

johanneswuerbach: No what I meant, wouldn't it be better to omit those properties when they are empty instead of setting them to an empty string?

calderas: It doesn't really make any difference for teamcity. It uses those fields ```message``` & ```details``` to provide extra information when inspecting a failed test. 

For example: 
* The first test on the image below shows what happens when those fields are empty.
* On the second test they are not empty. 
screen shot 2016-08-31 at 3 06 40 pm

Before this PR you don't get any details because the process exits with an error.
screen shot 2016-08-31 at 3 13 02 pm

PR #967, opened 2016-09-03
[FEATURE] Browser Flags (resolves #808)

### LINKED ISSUE #808: Add ability to pass flags to browser
It [looks like](https://github.com/testem/testem/blob/v1.4.1/lib/browser_launcher.js#L133) the command line arguments used for browsers are hard-coded. It would be nice to, for example, be able to pass the forthcoming [`--auto-open-devtools-for-tabs`](https://bugs.chromium.org/p/chromium/issues/detail?id=410958) to launch Chrome with DevTools opened.

jneurock:
Note: One test fails with Bluebird `3.4.6`. The test is in `/tests/ci/dev_tests.js` and the one that fails is `dev mode app > in interactive mode > allows to restart a run`.


johanneswuerbach:
This looks amazing, thanks for working on it.

jneurock:
Thanks for the kind words. Happy to work on it.

PR #974, opened 2016-09-08
Allow debug log file to be passed as option

### LINKED ISSUE #973: Configurable debug file location
Hi,

Would you consider the ability to configure the debug log location? We're currently experiencing sporadic timeouts on CircleCI and would like to save the debug file to the artifacts folder.

rupurt:
Fixes #973

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -62,8 +62,9 @@ Api.prototype.setup = function(mode, finalizer) {
 };
 
 Api.prototype.configureLogging = function() {
-  if (this.config.get('debug')) {
-    log.stream = fs.createWriteStream('testem.log');
+  var debug = this.config.get('debug');
+  if (debug) {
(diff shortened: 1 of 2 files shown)


lib/api.js
```diff
@@ -62,8 +62,9 @@ Api.prototype.setup = function(mode, finalizer) {
 };
 
 Api.prototype.configureLogging = function() {
-  if (this.config.get('debug')) {
-    log.stream = fs.createWriteStream('testem.log');
+  var debug = this.config.get("debug");
```
johanneswuerbach: Strings should be single quoted in this repo.

johanneswuerbach:
One lint issue otherwise lgtm

lib/api.js
rupurt: 👍 my bad :)

PR #984, opened 2016-09-19
Prevent slow script warning dialog in FireFox from hanging scripts.

adhamhf:
This PR will fix https://github.com/testem/testem/issues/953.

It sets the preference to stop the slow script dialog warning from appearing.


johanneswuerbach:
Instead of changing the default for everybody I would prefer a way to customize the path to the `prefs.js` so you can bring your own so we don't cause false-positives on actually super long running broken scripts.

adhamhf:
I can see that.  Its not clear to me how to do that and how to make ember do that though.

johanneswuerbach:
You can see nearly the same done for phantomjs here https://github.com/testem/testem/pull/982

> how to make ember do that though

ember wouldn't require any changes, just the `testem.js` inside your project would need to be changed to set this new config value to your custom `perfs.js`

adhamhf:
@johanneswuerbach this seems to work.  I've tested it locally, but don't know how to write a test for it.

adhamhf:
@johanneswuerbach this test appears to be failing on master too when I run them locally....

https://ci.appveyor.com/project/johanneswuerbach/testem/build/1428#L584

johanneswuerbach:
As I'm about to cut a new version I cherry-picked your commits and added a test: https://github.com/testem/testem/pull/987

I changed it to only allow to provide a custom `prefs.js` to not interfere with running multiple firefoxes in parallel (multiple profiles are required) and to prevent the provided profile path from being deleted.

adhamhf:
@johanneswuerbach this is almost entirely awesome.  One thing I learned when testing this is that you should never overwrite prefs.js.  Firefox states that you should use user.js for any custom settings.  Do you want me to open a new issue for that?  It thinks this would mean two things:

1.  Have the current settings written to user.js instead of prefs.js
2. Have users only be able specify a user.js instead of only a prefs.js.

johanneswuerbach:
Yes, sounds good to me. Could you share the source of this information? Is this backwards compatible?

adhamhf:
Its documented here: http://kb.mozillazine.org/User.js_file

One thing I notices is that if you commit prefs.js to git it changes every time you use the profile.  The user.js file does not appear unless you change it.

johanneswuerbach:
Sounds good, want to send a PR to rename it? After that I can push a new release.

adhamhf:
done, i think..  https://github.com/testem/testem/pull/989

PR #989, opened 2016-09-26
switching to user.js because Mozilla recommends that instead of prefs.js

adhamhf:
As per https://github.com/testem/testem/issues/988, this should allow testem to use user.js.

I'm not sure why but this test is failing, seems to fail in ci too

```
  ci mode app
    ✓ fails with explicitly defined missing launchers (97ms)
    ✓ passes when missing launchers are ignored (101ms)
    ✓ allows passing in reporter from config (97ms)
    ✓ wrapUp reports error to reporter (93ms)
    ✓ does not shadow EADDRINUSE errors
    ✓ stops the server if an error occurs (93ms)
not ok 1 TypeError
    ---
        message: >
            Cannot read property 'then' of undefined
    ...

1..1
# tests 1
# pass  0
# skip  0
# fail  1
```


tests/fixtures/firefox/custom_prefs.js
```diff
@@ -1,3 +0,0 @@
-user_pref("browser.shell.checkDefaultBrowser", false);
-user_pref("browser.cache.disk.smart_size.first_run", false);
-user_pref("dom.max_script_run_time", 0);
```
johanneswuerbach: Looks like this needs to be moved and not just deleted.

docs/config_file.md
```diff
@@ -70,7 +70,7 @@ Common Configuration Options
     config_dir:                  [Path]    directory to use as root for resolving configs, if different than cwd
     disable_watching:            [Boolean] disable any file watching
     fail_on_zero_tests:          [Boolean] whether process should exit with error status when no tests found
-    firefox_prefs_js:            [String]  path to firefox custom prefs.js file to be used
+    firefox_user_js:            [String]  path to firefox custom user.js file to be used
```
johanneswuerbach: One whitespace missing ;-)

adhamhf:
@johanneswuerbach push a fix.

johanneswuerbach:
Thanks, could you squash your commits?

adhamhf:
@johanneswuerbach squashed

PR #991, opened 2016-09-29
Namespace log events with console- prefix

davidhao3300:
Going to resolve the merge conflicts/clean up the code, we're on 1.6.0.

"error" is a reserved event in socket.io: socketio/socket.io#2285
Thus, console.error logs are never properly sent to the Testem server.
This namespaces console events so that we don't conflict with the reserved event.


johanneswuerbach:
Good catch :+1:

PR #1032, opened 2016-11-29
Fix a bug that console logs won't be shown in the terminal

svenpl:


johanneswuerbach:
Good catch, any idea how to add a test for this so we don't regress again?

Maybe something similar like this https://github.com/testem/testem/blob/master/tests/ci/ci_tests.js#L41, with just phantomjs and a fixture which logs some stuff.

PR #1034, opened 2016-12-01
move script tags to the body

stutrek:
I have a library that errors if it's in the head. I'm not interested in making it work in the head because the added complexity would have very little payoff.

This PR moves the script tags after the body in the templates. I'm open to making it a separate config to specify putting scripts either after the body or in the head.

Running `npm test` on master doesn't pass on my computer, as far as I can tell this PR does not make it any worse. `npm run integration` succeeds.


johanneswuerbach:
This feels a bit like a breaking change, would a custom template work for you?

stutrek:
For the vast majority of people it won't be a breaking change, you'd have to have a check for document.body === null, but I understand. I'd prefer a config to a custom template.

johanneswuerbach:
I'm fine with a new config option 👍

stutrek:
Ok, I've updated it with a new config option for footer_scripts.

```javascript
{
	"framework": "mocha+chai",
	"src_files": [],
	"footer_scripts": [
		"test/polyfills.js",
		"node_modules/sinon/pkg/sinon.js",
		"scrollMonitor.js",
		"test/tests.js"
	],

	"launch_in_dev": [
		"phantomjs"
	]
}
```

stutrek:
Side note, it seems that if you have Opera installed the tests will fail because it's using Opera as a launcher that's not installed.

stutrek:
@johanneswuerbach is this PR good? I'd like to use it in a second thing and I'd like to not use it from my fork

stefanpenner:
mind rebasing?

stutrek:
done

stefanpenner:
released as v2.8.0 🎉

PR #1037, opened 2016-12-06
Adding option to write launcher output to file.

paulot:
One common issue that we have at LinkedIn when using Testem is that we have no information on the underlying launcher that is propagated to the user. When running thousands of tests, it is not uncommon that browsers crash. This PR adds a flag "log_launchers" that when set to true will write the launcher's stdout and stderr to a file. This is particularly useful when things crash in ci, as we can have phantomjs' debug information propagated to the developer, rather than swallowed by Testem.


paulot:
@johanneswuerbach Any concerns over this?

johanneswuerbach:
I like the general idea, but would prefer to write the output into the testem debug log https://github.com/testem/testem/blob/241b1e42e96cd7e4d656efda6475e96e1f7978ae/testem.js#L18 as is already provides path customisation etc.

WDYT?

paulot:
Sounds good, I'm ok with that. How would it look in the testem log though? Would it all be outputted to the same log file?

johanneswuerbach:
I would use the debug flag without log launchers and always log in case of unexpected exit / error.

Would this help or do you need those logs when everything is fine?

johanneswuerbach:
For differentiation between launchers maybe a common prefix per launcher so you can easily cat log | grep launcherId ?

paulot:
Sounds good, logging onProcessError rather than onProcessExit is fine by me.

paulot:
Ok, it doesn't seem like onProcessError catches all of the error use cases that I'm looking for, so it doesn't seem to work. Any suggestions, or should I just keep it in onProcessExit?

johanneswuerbach:
It feels like you want to log in all cases where `toResult` is called https://github.com/testem/testem/blob/master/lib/runners/browser_test_runner.js#L79

Those are usually out-of-order exits when the process dies, the browser never connects or disconnects in the middle.

paulot:
Ok latest commit refactors the piece of code you pointed out and adds the logging functionality to the testem's log. Let me know if you think there are any additional changes.

lib/runners/browser_test_runner.js
```diff
@@ -143,6 +141,14 @@ BrowserTestRunner.prototype = {
     return this.launcher.name;
   },
 
+  reportResults: function(err, code, process) {
+    process = process || this.process;
```
johanneswuerbach: Could you name that differently? `process` is a global in node

johanneswuerbach:
Otherwise this looks pretty great, does that fix your debuggability issues?

lib/runners/browser_test_runner.js
paulot: Sure thing

paulot:
Yeah, this fixes the debuggability issues, hopefully we'll be able to root cause it now!

Thanks!

johanneswuerbach:
Perfect, could you squash your commits?

paulot:
Ok squashed and merged.

paulot:
@johanneswuerbach Changes are squashed, good to push?

johanneswuerbach:
Yes, sorry. Thanks for doing that 👍

Issue #803, opened 2016-02-25
XUnit output to stdout when xunit_intermediate_output and report_file are present?


rwjblue:
When the `xunit_intermediate_output` option is set to true and `report_file` is specified, I believe we should disable printing the XUnit output to `process.stdout`.  When both options are set, we actually get test output in three places (the report file, the intermediate TAP style output in stdout, and a giant blob of xml in stdout). 

This is currently done [here](https://github.com/testem/testem/blob/40d3c2afe898ff566cedfa209bd3354b708a82e8/lib/report_file.js#L16), and should be a relatively simple fix/change.

@johanneswuerbach - I wanted to run this by you to see if you would generally be :+1: or :-1:...

johanneswuerbach:
Yep, I think removing this makes totally sense :+1:

rwjblue:
Awesome, I'll work on that after getting my other PR to pass...

vvohra:
@johanneswuerbach Do you think this is still something do-able?

Since this issue was made the code has changed, so now a change is probably required somewhere [here](https://github.com/testem/testem/blob/master/lib/reporters/xunit_reporter.js#L39)

I'm wondering if I should try to make this generic, but I'm not sure if this is applicable to other reporters.

PR #1040, opened 2016-12-12
Suppressing xunit summary when using intermediate output

vvohra:
Fixes: #803

### DIFF
--- a/lib/reporters/xunit_reporter.js
+++ b/lib/reporters/xunit_reporter.js
@@ -36,8 +36,12 @@ XUnitReporter.prototype = {
       return;
     }
     this.endTime = new Date();
-    this.out.write(this.summaryDisplay());
-    this.out.write('\n');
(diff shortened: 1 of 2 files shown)


johanneswuerbach:
Great, could you add a test for this?

vvohra:
Added a test. Let me know if you want to change any wordings.

vvohra:
I realized that this disabled xunit output to stdout when intermediate is enabled. The original issue said to disable xunit output when intermediate is enabled AND a `report_file` is set. 

Thoughts if the PR should be updated?

johanneswuerbach:
I think this is fine, but maybe @rwjblue could share is use-case?

vvohra:
Hey, this isn't implemented correctly. I messed up and forgot to unit test for the XML output, I'm getting the XML output to be "Disabled XML to stdout when intermediate output is set" when it should have the results.

PR #1042, opened 2016-12-15
Suppress xunit with intermediate v2

vvohra:
Follow up to #1040

I'm not too happy with this. Because of the separation that exists now, this change is difficult to do. Before I write more tests I was wondering if you had some feedback.

Some alternatives:

- Pass a wrapper object to each Reporter that holds either a ReportFile or an out, the Reporter then can set properties on the ReportFile
- Let the reporter create the ReportFile and it can set the flag on its own
- `outputStream` can be piped an object instead of a string, the object would indicate if the string is meant for `out`, `fileStream` or both.

### DIFF
--- a/lib/reporters/xunit_reporter.js
+++ b/lib/reporters/xunit_reporter.js
@@ -36,12 +36,8 @@ XUnitReporter.prototype = {
       return;
     }
     this.endTime = new Date();
-    if (this.printIntermediateOutput) {
-      this.out.write('Disabled XML to stdout when intermediate output is set');
-    } else {
-      this.out.write(this.summaryDisplay());
-      this.out.write('\n');
-    }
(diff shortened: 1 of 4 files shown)


vvohra:
The other option that I might go with temporarily is just filter out `<testsuite></testsuite>` from the stdout when running this command.

johanneswuerbach:
Maybe the design should be change to create a dedicated reporter per output stream? This would also allow later to show the results to stdout with dot and write xunit to a file.

johanneswuerbach:
Implemented multiple reporters in https://github.com/testem/testem/issues/1044, would be cool if you could take a look.

vvohra:
Hey! I was on vacation. Looking through the changes now, but they seem good on first glance.

PR #1053, opened 2017-01-06
Extending middleware functionality, and documenting in config_file

### LINKED ISSUE #1052: Expose server to middleware
So that we can upgrade SSL proxying (like [this](https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/server/middleware/proxy-server/index.js#L31)) in testem middleware.

Would we be against passing [this.server](https://github.com/testem/testem/blob/fe38b04ca9c77e43122e8a0ec26ff28a7c9b7c69/lib/server/index.js#L98) to the [injectMiddlewareFunction](https://github.com/testem/testem/blob/fe38b04ca9c77e43122e8a0ec26ff28a7c9b7c69/lib/server/index.js#L155) ?

alexander-alvarez:
Closes https://github.com/testem/testem/issues/1052

### DIFF
--- a/tests/server_tests.js
+++ b/tests/server_tests.js
@@ -24,7 +24,7 @@ describe('Server', function() {
         port: port,
         src_files: [
           'web/hello.js',
-          { src: 'web/hello_tst.js', attrs: ['data-foo="true"', 'data-bar']}
+          {src:'web/hello_tst.js', attrs: ['data-foo="true"', 'data-bar']}
         ],
         routes: {
           '/direct-test': 'web/direct',
@@ -358,7 +358,7 @@ describe('Server', function() {
         cert: 'tests/fixtures/certs/localhost.cert',
         src_files: [
           'web/hello.js',
-          {src: 'web/hello_tst.js', attrs: ['data-foo="true"', 'data-bar']}
+          {src:'web/hello_tst.js', attrs: ['data-foo="true"', 'data-bar']}
         ],
         cwd: 'tests'
       });
--- a/tests/server_tests.js
+++ b/tests/server_tests.js
@@ -405,8 +405,8 @@ describe('Server', function() {
     });
   });
 
-  describe('injectMiddleware passes app and server', function() {
-    it(function(done) {
+  describe('injectMiddleware', function() {
+    it('passes app and server', function(done) {
       config = new Config('dev', {
         port: 0,
         cwd: 'tests',
--- a/tests/server_tests.js
+++ b/tests/server_tests.js
@@ -9,6 +9,8 @@ var fs = require('fs');
 var expect = require('chai').expect;
 var http = require('http');
 var https = require('https');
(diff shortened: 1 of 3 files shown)


johanneswuerbach:
Hey, could you add a test here https://github.com/testem/testem/blob/master/tests/server_tests.js to ensure we don't regress this feature in the future?

johanneswuerbach:
Friendly ping @alexander-alvarez :-)

alexander-alvarez:
:+1:  I'll get to it this long weekend

alexander-alvarez:
@johanneswuerbach  can you check that this is testing in the right way?

tests/server_tests.js
```diff
@@ -400,6 +404,33 @@ describe('Server', function() {
...
+      server = new Server(config);
+      server.once('server-start', function() {
+        done();
+      });
+      server.start();
+    });
+    after(function(done) {
+      server.stop(function() {
+        done();
+      });
```
johanneswuerbach: Could you move this after code into the test itself?

tests/server_tests.js
```diff
@@ -400,6 +404,33 @@ describe('Server', function() {
...
+          function(app, server) {
+            expect(app).to.be.instanceof(express);
+            expect(server).to.be.instanceof(BaseServer);
+            done();
+          }
+        ]
+      });
+      server = new Server(config);
+      server.once('server-start', function() {
+        done();
```
johanneswuerbach: Looks like this test uses `done` twice?

alexander-alvarez: modeled it after the `auto port assignment`, should I not do that?

johanneswuerbach: Yes, but the `done` function should generally only be called once in a test. In your test you call it twice, ones after the server is booted and once during the middleware creation. I guess the middleware one should be dropped. Could you make sure that it no longer passes when you stop passing `app` in the code?

alexander-alvarez:
>Could you make sure that it no longer passes when you stop passing app in the code?

@johanneswuerbach  not sure what you meant

tests/server_tests.js
```diff
@@ -400,6 +404,28 @@ describe('Server', function() {
       expect(config.get('port')).to.eq(server.server.address().port);
     });
   });
+
+  describe('injectMiddleware', function() {
+    it('passes app and server', function(done) {
+      config = new Config('dev', {
+        port: 0,
+        cwd: 'tests',
+        middlewear: [
```
johanneswuerbach: This property seems to contain a typo.

johanneswuerbach (changes requested):

NullVoxPopuli:
Would love to see some progress on this -- I'm trying to integrate testem with Vite, and it'd be a bit easier if I didn't need to figure out how to coordinate two servers :upside_down_face:

Issue #1056, opened 2017-01-11
`this.onFinish is not a function`


dfreeman:
Since updating to 1.14, I've started seeing `this.onFinish is not a function` errors from time to time stemming from `browser_test_runner.js`.

To be honest, I'm not familiar with the testem internals, and I'm a little murky on the exact sequence of events that's leading to this scenario, but I'm guessing that [this change](https://github.com/testem/testem/commit/2561a84b572fe8c82aeb214d721f4bca8315522e#diff-5eae7e485342946ae50eea2a59770c09R302) is likely the culprit. The call to `this.onFinish()` a few lines above is guarded on the truthiness of `onFinish`, and I'm guessing `this.exit()` is happy to receive an undefined callback, but the new `then()` handler just assumes it's defined.

Would just adding a guard there be reasonable, or is it expected that `onFinish` should always be defined at that point?

getsnoopy:
@dfreeman yes, that seems to be the issue. when i add a similar guard in the `return` statement, the error goes away.

johanneswuerbach:
Sounds good, would happily accept a PR :-)

PR #1057, opened 2017-01-11
Guard this.onFinish() in browser_test_runner.js

dfreeman:
This change adds a simple guard to avoid calling `onFinish` if it's not defined, fixing #1056.

Not sure how(/whether) to test this — any guidance would be appreciated 🙂

### DIFF
--- a/lib/runners/browser_test_runner.js
+++ b/lib/runners/browser_test_runner.js
@@ -264,7 +264,10 @@ BrowserTestRunner.prototype = {
       return;
     }
     return this.exit().then(function() {
-      this.onFinish();
+      // TODO: Not sure how this can happen, but sometimes onFinish is not defined
+      if (this.onFinish) {
+        this.onFinish();
+      }
     }.bind(this));
   }
 };


johanneswuerbach:
Also not really sure how this can happen, fine with this hack for now.

Could you add a comment like: `// TODO: Not sure how this can happen, but sometimes onFinish is not defined`?

dfreeman:
Will do. The only scenario where I could come up with for it to be happening in is if `finish()` gets called but not `start()`, which doesn't really make sense to me. Given that the other `this.onFinish()` call is similarly guarded, though, I'm guessing this isn't a new thing ¯\\\_(ツ)_/¯

getsnoopy:
this happens whenever i initially access the Testem server in Chrome (after it finishes running the tests). when i try the same now, it doesn't throw the error

getsnoopy:
thanks for merging it in! could you bump the version version as well?

johanneswuerbach:
Sorry, released as `v1.14.3`.

PR #1060, opened 2017-01-17
Support running tests from a file: URL

bendemboski:
## Disclaimer

I'm pretty new to the testem code base, so I'm opening this up to see if it or something like it makes any sense, and to get feedback. Maybe there's an easier way of getting electron to run under testem, and if so I'd love to hear it, but this is the best I could figure out. Also, I wrote an integration test, but couldn't really find where unit tests related to these changes live, so I'd love a pointer there.

## The proposed change

There are a couple of changes here that enable (with some hoop-jumping) running tests from file: urls. My use case is electron -- testem is awesome, and I want to be able to test my electron apps using testem. More specifically, I'm doing some work on ember-electron, and ember is already set up to use testem. However, as you can see from the integration test, these changes will support electron in general, not just ember electron.

The primary changes are:
* Instead of using the URL `/testem/connection.html` to load the client into the iframe, parse out testem.js' script tag's src and construct a fully-qualified URL including the scheme, so that the URL from which testem.js is loaded determines where to find connection.html, not the URL from which the host HTML is loaded.
* Allow the host HTML to specify a global function to do custom discovery of the testem ID, so that it can be provided through some other mechanism than being the first path component.

With these two changes in place, a custom runner can process the testem-generated URL, rewrite it to point into the filesystem, munge the test HTML a bit to point to testem.js with a fully-qualified URL, move the testem ID to a query param so it doesn't mess up the file: URL, and then run HTML with an extra <script> tag providing a global function to parse the testem ID out of the query param instead of looking in the path. Bing bang boom, electron runs all the code out of the filesystem, but reports everything seemlessly to testem via the browser protocol as if it were an actual browser running code through the testem proxy server.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ os:
 - osx
 - linux
 osx_image: xcode8
+addons:
+  apt:
+    packages:
+      - xvfb
 language: node_js
 node_js:
   - "0.10"
@@ -27,6 +31,8 @@ before_install:
     fi
   - npm config set spin false
 install:
+  - export DISPLAY=':99.0'
+  - Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
   - travis_retry npm install
 before_script:
   - if [ $TRAVIS_OS_NAME == "linux" ] && [ $TRAVIS_NODE_VERSION == "4" ]; then
--- a/bin/run-integration.js
+++ b/bin/run-integration.js
@@ -28,7 +28,8 @@ var skipOnWindows = [
 ];
 var skipDefiningReporter = [
   'node_example',
-  'node_tap_example'
+  'node_tap_example',
+  'electron'
 ];
 var examplesPath = path.join(__dirname, '../examples');
 var DEFAULT_CONCURRENY = 10;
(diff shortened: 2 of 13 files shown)


bendemboski:
I can debug what's going wrong with the integration test in the next day or two (I've only tested locally on MacOS), but in the meantime I'd love some general feedback on this direction so I don't waste my time if there's an entirely better way to do this all.

johanneswuerbach:
👏  the general approach looks good 👍  Looks like all failing tests are trying to run your electron example on linux, that might be the culprit.

As a minor nit, could you change the example to include expect.js via npm instead of vendoring it? That should also remove 1.2k of the lines added in this PR.

bendemboski:
@johanneswuerbach I think this is ready to go. I removed the extra files (that was actually me carelessly commiting a directory that I didn't mean to), bound a few more variables into the template to simplify the runner script, added unit test coverage for them, and fixed the integration tests.

I haven't rebased so you can see what changed, but I think this is logically just one commit, so I'm happy to rebase/squash if you give the word, or you can do so when merging. Let me know what you prefer, and if you see any other issues for me to address!

johanneswuerbach:
This looks really great 👏 Could you rebase? Then this should be ready to be merged :-)

bendemboski:
rebased!

PR #1070, opened 2017-02-07
Provide testem's cwd to launcher templates

bendemboski:
This is a follow-up to #1060 -- I think of it at the time, but often systems using file-url functionality will be running on built/minified/etc output in a temp directory. Testem makes this easy with the `cwd` options, telling it where to serve files from. Propagating the `cwd` into the template context allows test runners to know where the built content is so they can mirror this behavior.


johanneswuerbach:
Makes sense, thanks :clap:

PR #1071, opened 2017-02-09
Fix multiple tab/reporter issue

### LINKED ISSUE #1063: Many chrome browsers start
I'm using testem with and ember-cli project in dev mode launching Chrome. 

When I'm pausing at breakpoints to debug and saving files as I fix things, testem will launch new instances of Chrome. Before long I'll have five instances of Chrome running, all running my tests and slowing down my machine and hogging my CPU. Is this expected behavior or have I configured something wrong? or is it a bug. I'm happy to poke around, but I'm not sure where I would begin.

bendemboski:
When running in server mode, if something causes the browser to reload (e.g. the user changing the QUnit filters to run a subset of the tests, or a live reload in a framework like ember-cli), a new reporter and UI tab will be spawned. This was happening because the browser login code was only matching against still-pending runners.

We can make the check more resilient by instead of checking pending, just looking for a runner with the correct id that either doesn't have a socket, or whose socket is not connected. That allows the same browser to connect back to the same reporter any number of times across page refreshes.

This may also fix #1063 although I haven't been able to reproduce that issue myself, so I can't be sure if this fixes it or not.

I'm not really sure how to unit test this fix -- I'm open to suggestions :)

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -233,7 +233,7 @@ App.prototype = {
 
   onBrowserLogin: function(browserName, id, socket) {
     var browser = find(this.runners, function(runner) {
-      return runner.pending && runner.launcherId === id;
+      return runner.launcherId === id && (!runner.socket || !runner.socket.connected);
     });
 
     if (!browser) {
--- a/tests/ci/dev_tests.js
+++ b/tests/ci/dev_tests.js
@@ -3,6 +3,7 @@
 var fs = require('fs');
 var expect = require('chai').expect;
 var path = require('path');
+var sinon = require('sinon');
 
 var App = require('../../lib/app');
 var Config = require('../../lib/config');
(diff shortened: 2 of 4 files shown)


johanneswuerbach:
Nice research. I think adding a test similar to https://github.com/testem/testem/blob/bc28d03ffcd27fbe681a367ab59a4ac61a9cf382/tests/ci/ci_tests.js#L165-L182

Instead of redirect to a different page https://github.com/testem/testem/blob/bc28d03ffcd27fbe681a367ab59a4ac61a9cf382/tests/fixtures/disconnect-test/tests.js#L11, the fixture could redirect to testem itself with a query parameter when none is defined and just finish the tests when the parameter is set. WDYT?

mfeckie:
I can confirm that this works nicely for scenarios where you're just applying filters.  Still have issues with additional instance being fired up if using the debugger.

I suspect there's some timeout code that aims to recover if a session become unresponsive / takes too long.

johanneswuerbach:
@mfeckie to be really sure, could you try to increase your `browser_disconnect_timeout` https://github.com/testem/testem/blob/master/lib/runners/browser_test_runner.js#L228 to something really high?

I guess debugging pauses any js and also kills somehow influences the websocket connection.

bendemboski:
@johanneswuerbach unit test is implemented. I think this is ready to go if you're happy with it.

PR #1072, opened 2017-02-14
Mde/fix console events

zen-mde:
Emits an event when the Socket.io server is created.


zen-mde:
Should have raised against zenefits-release, closing.

PR #1080, opened 2017-03-07
move to proxy behind the static files, so you serve static files before the proxies

albertjan:
I wanted to proxy a website into an iframe to run tests on the website. But for that to work the proxy url needed to be `/`. So all the assets can come in via the proxy aswel.

This however creates a problem because the proxy then intercepts all requests to the testem files and my local tests files. 

To solve this I moved the proxy to after the `serveStaticFile` calls. To make sure the serveStaticFile doesn't return a 404 but continues on to the next route I've added a call to `next('route')` in the `serveStaticFile` function.

I've had to change only one failing tests: `'proxies get html request to api3'` which is probably the wrong name for the tests as it doesn't test the proxy but the error handler. 

This change does cause that to fail, because when the file is not found it tries something else instead of throwing an error. 

To solve this I think we could just add `app.all(".*", (req, res) => res.send("Not found: " + req.originalUrl));` at the end to catch everything.

### DIFF
--- a/lib/server/index.js
+++ b/lib/server/index.js
@@ -111,8 +111,6 @@ class Server extends EventEmitter {
 
     this.injectMiddleware(app);
 
-    this.configureProxy(app);
-
     app.get('/', (req, res) => {
       res.redirect(`/${String(Math.floor(Math.random() * 10000))}`);
     });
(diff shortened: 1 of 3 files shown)


johanneswuerbach:
Thanks, looks like a nice great addition.

> To solve this I think we could just add app.all(".*", (req, res) => res.send("Not found: " + req.originalUrl)); at the end to catch everything.

Sounds great, would you mind adding that.

Could you also add a tests for the new behaviour so we don't regress?

albertjan:
@johanneswuerbach better late then never right 😄 . Sorry to have let this sit for so long.

PR #1083, opened 2017-03-10
tap_quiet_logs: only show logs for failed tests

krallin:
This adds a tap_quiet_logs option to only show logs when a test fails.
The option defaults to false, which is the current behavior (i.e. show
logs whenever they're present).

The underlying motivation here is to be able to silence verbose logging
we get when receiving our tests, except when the tests fail, so that we
can have generally quiet logs and full debug information if a test
fails.


johanneswuerbach:
Thanks, this looks great.

rwjblue:
@johanneswuerbach - Any chance we could get a release including this?

johanneswuerbach:
Sorry, I'll create a new release later today. (also open to provide that permissions to more people if interested)

PR #1090, opened 2017-03-16
More thorough description of `data` provided to `report`

brettz9:
I thought these ought to be better documented. This is just my understanding from looking around a bit, so I'd appreciate confirmation.


johanneswuerbach:
Looks correct, but will double check when I've some time.

johanneswuerbach:
This looks good 👍 Thanks again 👏

PR #1100, opened 2017-03-23
Testing and known browsers

brettz9:
- Only add "Safari Technology Preview" to known browsers if not on Windows
- Testing: Prevent potential `beforeEach` error
- Testing: Suppress test output for "kills launchers on wrapUp" ci test
- Testing: Instead of assuming opera is missing, assume IE is missing on non-Windows and assume Safarai Technology Preview is missing on Windows


johanneswuerbach:
Good idea :+1:

Issue #1115, opened 2017-04-27
Can no longer inherit from internal reporters.


rwjblue:
As of https://github.com/testem/testem/pull/1083, the built-in reporters get an additional argument (the `config` object) passed into them. Prior to #1083 it was possible to extend one of the default reporters:

```js
// testem.js
const TapReporter = require('testem/lib/reporters').tap;
class FailureOnlyReporter extends TapReporter {
  constructor() {
    super(...arguments);
    this._reportCount = 0;
  }

  display(prefix, result) {
    this._reportCount++;

    if (!result.passed) {
      super(...arguments);
    }

    if (this._reportCount > 100) {
      this.out.write('pass count: ' + this.pass);
      this._reportCount = 0;
    }
  }
}

module.exports = {
  // ...snip...
  reporter: new FailureOnlyReporter()
}
```

(we do this in Travis for Ember because otherwise the logs end up being longer than the CI output allows and the builds are failed)

After #1083, it is no longer possible, because the `config` object itself is not available when we are instantiating the reporter. 

---

One possible solution would be to modify [this code](https://github.com/testem/testem/blob/master/lib/utils/reporter.js#L12-L19) to also check if the name is a function and if it is, then instantiate it in the same way that a string version is.

This would allow us to use:

```js
module.exports = {
  //...snip...
  reporter: FailureOnlyReporter
};
```

And Testem would instantiate itself...

johanneswuerbach:
Sounds like a good fix before rethinking this for a 2.0. I'm currently a bit busy, would you have time to create a PR for this?

PR #1148, opened 2017-06-27
Inheriting from internal reporters

trjohnst:
(This is my first pull request so please let me know if I'm missing anything major)

This is an attempt to re-enable extending off of custom reporters based off @rwjblue's thoughts in #1115. Please see more details below:

# Problem Statement
The [currently documented route](https://github.com/testem/testem/blob/master/docs/custom_reporter.md#configuration) of passing `reporter: new MyCustomReporter()` no longer works for extending off of internal reporters such as the [TapReporter](https://github.com/testem/testem/blob/master/lib/reporters/tap_reporter.js) due to [how it accesses config options](https://github.com/testem/testem/blob/master/lib/reporters/tap_reporter.js#L8) introduced by 3729d91 to support configuration for the [TapReporter](https://github.com/testem/testem/blob/master/lib/reporters/tap_reporter.js). 

As a means of re-enabling usage of configs and custom reporters, passing a constructor reference instead of a reporter instance was suggested in #1115 as a means of keeping config management internal.

# Summary of Changes
- Test case to capture the use case (perhaps there's a better place or more tests to write, please let me know)
- Code to handle passing in a constructor reference to a testem configuration
- Documentation updates

### DIFF
--- a/docs/custom_reporter.md
+++ b/docs/custom_reporter.md
@@ -71,8 +71,5 @@ To use your custom reporter, set `reporter` in your `testem.js` config file:
             "src/*.js",
             "tests/*_tests.js"
         ]
-        "reporter": new MyReporter()
+        "reporter": MyReporter
     };
-
-
-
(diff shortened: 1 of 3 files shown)


johanneswuerbach:
LGTM 👍

PR #1155, opened 2017-07-13
QUnit Adapter: Stop manually tracking currentModule and use testStart params instead

trentmwillis:
This was causing a bug with the `random` feature in QUnit (see https://github.com/trentmwillis/ember-exam/issues/99 for more info). Additionally, `QUnit.moduleStart` was being improperly set here.


johanneswuerbach:
Good catch :clap: Thanks!

Issue #1125, opened 2017-05-12
on_exit doesn't run if error occurs before it


getsnoopy:
If an error occurs sometime in the lifecycle before running the `on_exit` hook, the `on_exit` hook doesn't run.

For example, it seems like PhantomJS is terminating with exit code `1`, but there's no `Stopping ci` message in the debug logs:
```
info Test'em starting..
info Seeking for config file...
info Starting ci
info Starting server
info Waiting for tests.
info Start triggered test run.
info Running tests...
info Tests running.
info spawning: phantomjs - [ '/Users/getsnoopy/inkwell/node_modules/testem/assets/phantom.js',
  'http://localhost:7357/784/.tools/test/unit/index.mustache' ]
info New client connected: PhantomJS 2.1 784
info tryAttach PhantomJS 2.1 784
info Browser PhantomJS finished all tests. true
info Closing browser PhantomJS.
WARN PhantomJS closed 1
info Process PhantomJS terminated. 1 null
```

PR #1157, opened 2017-07-21
Call on_exit hook even on failure

getsnoopy:
Previously, the on_exit hook was only being called when all the
tests have passed without errors. This leads to side effects, so
this commit fixes this issue (#1125).


johanneswuerbach:
Good catch :clap:

Issue #1122, opened 2017-05-06
The 'cwd' config parameter is not respected when determining available launchers


getsnoopy:
The `cwd` config parameter allows changing the current working directory, which is not used when determining available launchers (by way of adding `<project root>/node_modules/.bin` to the `PATH`).

So this doesn't work e.g., when PhantomJS is installed locally, but the CWD needs to be changed.

The issue is in `lib/env-with-local-path.js`:
```
var modulesPath = path.join(process.cwd(), 'node_modules', '.bin');
```

PR #1158, opened 2017-07-22
Use config.cwd() to determine available launchers

getsnoopy:
The "cwd" property from the config was not respected when
determining available launchers. If the current working directory
was different than the one from the current process, the available
launchers could not be accurately determined. This commit fixes
this issue (#1122).


johanneswuerbach:
Thanks :+1:

Issue #1129, opened 2017-05-24
Mocha test logs twice when failed


CodeDeFury:
On latest version of testem, for this example: [examples/mocha_simple](https://github.com/testem/testem/tree/master/examples/mocha_simple), make the test to fail and i get two logs on the failing test as follow:

```
> node ../../testem.js ci -P 10

not ok 1 Safari 10.1 - hello should say hello
    ---
        message: >
            expected 'hello world' to equal 'hello world1'
        stack: >
            assert@expect.js:99:22
            equal@expect.js:200:16
            expect.js:73:29
            hello_spec.js:3:30
        Log: |
    ...
not ok 2 Safari 10.1 - hello should say hello
    ---
        message: >
            expected 'hello world' to equal 'hello world1'
        stack: >
            assert@expect.js:99:22
            equal@expect.js:200:16
            expect.js:73:29
            hello_spec.js:3:30
        Log: |
    ...
not ok 3 PhantomJS 2.1 - hello should say hello
    ---
        message: >
            expected 'hello world' to equal 'hello world1'
        stack: >
            assert@http://localhost:7357/expect.js:99:27
            equal@http://localhost:7357/expect.js:200:16
            http://localhost:7357/expect.js:73:29
```

getsnoopy:
This issue should be fixed now, as of #1167.

PR #1045, opened 2016-12-20
Add tests and fix bug in mocha_adapter

job13er:
Fixes #1043

Issue #1043, opened 2016-12-16
Errors in beforeEach not reported in ci mode in mocha


job13er:
With the most recent versions of `mocha`, the `test end` event is no longer emitted when there is an error thrown in a `beforeEach()`, the result is that in CI mode. Previously the `test end` event was emitted whenever there's a `fail` event, but that's no longer the case. As a result, `testem` no longer shows the error. This manifested for my team when using `ember-cli` and `ember-cli-mocha`.

I was able to reproduce with a very simple test: 
```js
/**
 * This is a test that forces a failure in a beforeEach to make sure
 * that it is properly reported. We were seeing an issue with the latest `ember-mocha` where
 * testem wasn't reporting errors in this case
 */

import {expect} from 'chai'
import {beforeEach, describe, it} from 'mocha'

describe('Outer test that passes', function () {
  let foo
  beforeEach(function () {
    foo = 1
  })

  it('should fail', function () {
    expect(foo).to.equal(1)
  })

  describe('Inner test with bad beforeEach', function () {
    beforeEach(function () {
      throw new Error('error in beforeEach')
    })

    it('should have failed before it got here', function () {
      expect(foo).to.equal(1)
    })
  })
})
```

When running in the browser (dev mode) I see

![browser-error](https://cloud.githubusercontent.com/assets/411788/21279827/908c6f14-c39f-11e6-9686-f8979c61be7a.png)

However, the console output is: 

![console-no-error](https://cloud.githubusercontent.com/assets/411788/21279854/bb0041f8-c39f-11e6-9961-eb60bbce9f55.png)

And when running in CI mode, I get:

```
ok 1 Chrome 55.0 - Outer test that passes should fail

1..1
# tests 1
# pass  1
# skip  0
# fail  0

# ok
```

I locally tried updating the `mocha_adapter` [here](https://github.com/testem/testem/blob/master/public/testem/mocha_adapter.js#L72) to do: 
```js
var name = getFullName(test);
test.err = test.err || err;
testFail(test, err);
```
instead of just 
```js
test.err = test.err || err;
```

And the result was promising. I then saw

![console-error](https://cloud.githubusercontent.com/assets/411788/21280021/c7c7faa6-c3a0-11e6-8781-18df24055a20.png)

in dev mode and for ci mode I got:

```
not ok 1 Chrome 55.0 - Outer test that passes Inner test with bad beforeEach "before each" hook for "should have failed before it got here"
    ---
        message: >
            error in beforeEach
        stack: >
[...]

PR #1165, opened 2017-08-09
Fix Mocha's duplicate failure logs

getsnoopy:
Previously, there was a bug in Mocha where the "fail" event would
not call the "test end" event, resulting in some errors not being
reported. This was accounted for on the Testem side by #1045, but
Mocha has since fixed this issue. This resulted in a bug where
test failures were being reported twice, so this commit fixes
the issue (#1129) by removing the original workaround.

### DIFF
--- a/public/testem/mocha_adapter.js
+++ b/public/testem/mocha_adapter.js
@@ -70,8 +70,6 @@ function mochaAdapter() {
           emit('all-test-results', results);
         }
       }, 0);
-    } else if (evt === 'fail') {
-      testFail(test, err);
     }
 
     oEmit.apply(this, arguments);
--- a/tests/mocha_adapter_tests.js
+++ b/tests/mocha_adapter_tests.js
@@ -402,21 +402,8 @@ describe('mochaAdapter', function() {
         expect(originalEmit).to.have.been.calledWith(evt, test, err);
       });
 
-      it('should emit a "test-result" event', function() {
-        expect(_emit).to.have.been.calledWith('test-result', {
-          failed: 1,
-          id: 1,
-          items: [{
-            passed: false,
-            message: 'The error message',
-            stack: 'The stack trace'
-          }],
-          name: 'foo ',
-          passed: 0,
-          pending: 0,
-          runDuration: 123,
-          total: 1
-        });
+      it('should not emit a "test-result" event', function() {
+        expect(_emit).not.to.have.been.called();
       });
     });
   });


johanneswuerbach:
👍

job13er:
It would have been really nice to include a reference to the commit/version in `mocha` where this is fixed, since at least some of our setups are now pulling in the still broken `mocha` version and this now incompatible `testem` version, regressing #1043 from out point of view.

job13er:
The most recent reference to `test end` in `mocha`'s `CHANGELOG.md` seems to indicate that emitting `test end` in `beforeEach` is a bug, not a feature:
https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#050--2011-12-14

I guess I'll keep looking for the "fix" that starts emitting it again.

johanneswuerbach:
Sorry, I'm a bit lost now. This this be reverted or now?

job13er:
Sorry, let me try to explain.

While the latest `mocha` may have started sending `test end` when there's an error in `beforeEach`, `mocha@2.5.3` still does not. 

So, this change broke functionality for anyone still using `mocha@2`. I don't know exactly when mocha started sending `test end` on exceptions in `beforeEach`, so I don't know what minimum version of `mocha` is required for this fix to really be a fix, and not a regression. My first comment was expressing my frustration that this PR didn't reference *when* "Mocha has since fixed this issue" 

When I started looking through `mocha`'s `CHANGELOG.md` to find the fix referenced here, I found only two mentions of "test end", which seemed to indicate that it was intentional behavior for `mocha` to *not* send `test end` when there is a failure in a hook. That confused me quite a bit. 

The end result is that this change, which was released as a PATCH, broke integration between `testem` and `mocha@2.5` in a way that's quite annoying because it lets test failures go unnoticed and broken code get merged. 

I'm pinning all our projects to use `testem@1.18.1` until we can figure out if we can move forward to whatever version of `mocha` works properly with `testem@1.18.2`. Sadly, `ember-cli-mocha` is still on `mocha@^2.5.0` so that's what we're stuck with at least in the short term. 

So, if you want `testem>=1.18.2` to work with `mocha@2.5` then something needs to be done. If you don't want it to work with that `mocha` version, then I suppose you don't have to do anything.

johanneswuerbach:
Sounds like a revert then for me, WDYT @getsnoopy ?

getsnoopy:
I see the issue now. Yes, let's revert this one. I think I know how to fix it so that it fixes the issue without causing a regression with error reporting. I'll put up a PR soon.

johanneswuerbach:
Thank you both, reverted and published as `v1.18.3`. @job13er sorry about the regression, this wasn't intended.

PR #1167, opened 2017-08-12
Fix Mocha's duplicate error messages on failure

getsnoopy:
Mocha fires a "fail" event for test failures and hook failures, but
fires a "test end" event only for tests. Since we want to handle
hook failures as well, the "fail" event was accounted for in #1045.
But this caused error messages to be logged twice since the handler
code was also firing for the corresonding "test end" events that
are fired for every "fail" event. This commit removes the handler
code from the "test end" event for failed tests so that it logs
errors only once.

### DIFF
--- a/public/testem/mocha_adapter.js
+++ b/public/testem/mocha_adapter.js
@@ -61,8 +61,6 @@ function mochaAdapter() {
         waiting--;
         if (test.state === 'passed') {
           testPass(test);
-        } else if (test.state === 'failed') {
-          testFail(test, err);
         } else if (test.pending) {
           testPending(test);
         }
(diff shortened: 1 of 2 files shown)


getsnoopy:
This should properly fix #1129 without causing a regression for #1043. @job13er, can you verify?

job13er:
Checking now

job13er:
@getsnoopy It appears that the use-case where `testem@1.18.2` was not displaying an error in `beforeEach` for us is still correctly displaying the error when using your branch here. So I think this is good to go. 

👍

johanneswuerbach:
Sorry, thanks for the bump.

PR #1170, opened 2017-08-23
Improve the output of the dot reporter

smottt:
A small improvement to the output of the dot reporter. At the moment it just displays dots indefinitely which is not pleasant to see if the terminal is wide. This PR limits the output to 60 dots per line. 60 is just an arbitrary number that can be adjusted if needed. It also makes the report of failed tests more readable.

Before this PR (this could just be one straight line if the terminal is wide enough):
before

After this PR:
after

### DIFF
--- a/lib/reporters/dot_reporter.js
+++ b/lib/reporters/dot_reporter.js
@@ -84,7 +84,7 @@ DotReporter.prototype = {
         printf(this.out, '\n%s', indent(error.stack, 5));
       }
 
-      this.out.write('\n');
+      this.out.write('\n\n');
     }, this);
   },
   summaryDisplay: function() {
--- a/lib/reporters/dot_reporter.js
+++ b/lib/reporters/dot_reporter.js
@@ -14,6 +14,7 @@ function DotReporter(silent, out) {
   this.results = [];
   this.startTime = new Date();
   this.endTime = null;
+  this.currentLineChars = 0;
   this.out.write('\n');
   this.out.write('  ');
 }
@@ -36,6 +37,10 @@ DotReporter.prototype = {
     if (this.silent) {
       return;
     }
+    if (this.currentLineChars > 60) {
+      this.currentLineChars = 0;
+      this.out.write('\n  ');
+    }
     if (result.passed) {
       this.out.write('.');
     } else if (result.skipped) {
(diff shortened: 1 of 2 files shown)


lib/reporters/dot_reporter.js
```diff
@@ -36,13 +37,18 @@ DotReporter.prototype = {
     if (this.silent) {
       return;
     }
+    if (this.currentLineChars > 60) {
```
johanneswuerbach: Could you maybe use https://nodejs.org/api/tty.html#tty_writestream_columns to detect length from the terminal?

smottt: Definitely! I'll give it a shot and let you know.

johanneswuerbach: @smottt did you had the chance to take a look?

smottt: @johanneswuerbach Sorry about the long delay, was quite busy lately. I'll check this at some point this week.

smottt:
I updated the code as suggested. I decided to still limit to max 60 characters. Those `-5` are there to compensate for the spaces around the line. One thing I'm unsure of is whether the `columns` property is always present and set on `this.out`?

Please review the code and let me know if you wish to have any other modifications done.

johanneswuerbach:
I think the `columns` property isn't always present, lets default that to `60` and then this should be good to go 👍

smottt:
It actually is 60 in the end, due to that `- 5` which is necessary to adjust the paddings around it. Unless you want a number 60 there which would result in max dots of 55?

stefanpenner:
@johanneswuerbach ping, any changes to be done still?

tandrewnichols:
Nothing personal, but I couldn't dislike this change more. Can I override it?

tandrewnichols:
I should clarify my negative feedback. There's two reasons I don't view this as an improvement: 1) Maybe it looks fine in your example with 418 tests, but we have over 3000 and now they no longer all fit on the screen at once. 2) The full width is sort of the standard dot reporting format. We run multiple test suites in parallel and the other dot reporters _are_ printing full width so this looks weird, and of course, it _even more_ doesn't fit on the screen when there are other test suites running too.

stefanpenner:
@tandrewnichols I would be very open to accepting a PR to make this configurable :)

tandrewnichols:
Cool, I can probably swing that.

PR #1193, opened 2017-11-04
Display error message correctly when a hook fails to run.

### LINKED ISSUE #1192: Testem shows undefined error when before_tests fails
See https://github.com/tjgq/testem-browserify-error, minimally adapted from https://github.com/testem/testem/tree/master/examples/browserify. There is a syntax error in `hello.js`, which causes the browserify command to fail. When I run testem with `npm test`, instead of the browserify error message, the console UI shows
```
Testem error

undefined
undefined
```

The problem is in `lib/reporters/dev/index.js`:
```
  report: function(browserName, result) {
    if (isTestemItself(result)) {
      return this._showError('Testem error', result.error.message);
    }
```

The `_showError` method should be called with `result.error` rather than `result.error.message`.

tjgq:
Fixes #1192.

### DIFF
--- a/lib/reporters/dev/index.js
+++ b/lib/reporters/dev/index.js
@@ -336,7 +336,7 @@ module.exports = View.extend({
 
   report: function(browserName, result) {
     if (isTestemItself(result)) {
-      return this._showError('Testem error', result.error.message);
+      return this._showError('Testem error', result.error);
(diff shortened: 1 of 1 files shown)


johanneswuerbach:
Thank you 👍

PR #1195, opened 2017-11-20
fix latest chrome exist on linux with version 0.8.5

vincent178:
with latest chrome 62.0.3202.94-1 installation on ubuntu, chrome exist directly.
add "--no-sandbox" to fix this issue on v.0.8.5.

### DIFF
--- a/lib/browser_launcher.js
+++ b/lib/browser_launcher.js
@@ -217,7 +217,7 @@ function browsersForPlatform(){
         name: 'Chrome',
         exe: 'google-chrome',
         args: ["--user-data-dir=" + tempDir + "/testem.chrome",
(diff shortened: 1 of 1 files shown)


vincent178:
This travis node version seems pretty old, I think it would be good if node version is 6.x or after.

stefanpenner:
> "--no-sandbox"

I'm not why this is required, can you share more details?

Also the code has changed, it now resides here: https://github.com/testem/testem/blob/dca112ff576d2c66f7cca3d091936ddd7acd3084/lib/utils/known-browsers.js#L24
---


Closing, if rationale for ^ is provided, and code is upgraded we can consider a new PR

PR #1204, opened 2018-01-19
[BUGFIX] Prevent leaked runners/sub processes.

stefanpenner:
It is not uncommon during developer to end the process unexpectedly. Examples are `<ctr-c>`, crash, etc.

Switching to `execa` gives us

* automatic cleanup of child_processes in the above scenarios
* Promise API
* other nice things: https://github.com/sindresorhus/execa


stefanpenner:
hmmm.... well apparently I just trolled myself... time to fix...

PR #1205, opened 2018-01-19
[BUGFIX] Prevent leaked runners/sub processes.

stefanpenner:
It is not uncommon during developer to end the process unexpectedly. Examples are <ctr-c>, crash, etc.

Switching to `execa` gives us

* automatic cleanup of child_processes in the above scenarios
* Promise API
* other nice things: https://github.com/sindresorhus/execa#why

### DIFF
--- a/bin/run-tests.js
+++ b/bin/run-tests.js
@@ -1,39 +1,21 @@
 #!/usr/bin/env node
 'use strict';
 
-var spawn = require('child_process').spawn;
+var execa = require('execa');
+var command = 'npm';
 
-function run(command, args, cb) {
-
-  console.log('Running: ' + command + ' ' + args.join(' '));
-
-  var child = spawn(command, args, { stdio: 'inherit' });
-
-  child.on('error', function(err) {
-    cb(err);
-  });
-
-  child.on('exit', function(code) {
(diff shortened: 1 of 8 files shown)


stefanpenner:
@rwjblue  / @krisselden r?

johanneswuerbach:
Looks reasonable, could you remove the `yarn.lock`? Otherwise 👍

stefanpenner:
The remaining Travis Task was OSX, which is having some trouble. Prior to dropping the lock file it was green, and I have now been waiting 1+ day for the job to be executed.

Local test suggest all is well, as was the last CI run... So I am making the judgement call to skip it.

stefanpenner:
@johanneswuerbach I believe we must release 2.x correct? Should I just do that now?

johanneswuerbach:
@stefanpenner done, support about the delay.

stefanpenner:
@johanneswuerbach nice thanks!

stefanpenner:
@johanneswuerbach now that we support only newer node, any opposition to me modernizing the codebase?

johanneswuerbach:
Definitely not! Happy to review.

PR #1208, opened 2018-01-24
Make the progress indicator useful for QUnit

stefanpenner:
Test output will not contain:

```
numberOfTestsThatHaveRun/numberOfTestsThatShouldRun
```

instead of

```
numberOfTestsThatHaveRun/numberOfTestsThatHaveRun
```

----

notes:
* Depends on: https://github.com/qunitjs/qunit/pull/1256
* Has the current behavior as a fallback

### DIFF
--- a/public/testem/qunit_adapter.js
+++ b/public/testem/qunit_adapter.js
@@ -28,6 +28,7 @@ function qunitAdapter() {
     tests: []
   };
   var currentTest;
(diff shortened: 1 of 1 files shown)


stefanpenner:
Apparently their is a better API for us to use [outlined here](https://github.com/qunitjs/qunit/pull/1256)

PR #1209, opened 2018-02-05
UTF8 Validation TypeError

LoicEvenium:
Update socket-io to fix "UTF8 Validation TypeError" from ws. See https://github.com/websockets/ws/issues/1021


johanneswuerbach:
@stefanpenner I’m not sure about this one as far as I know socket.io 1.6 was the last release without native dependencies and every newer release has again optional native dependencies or did our opinion regarding those changed?

stefanpenner:
@johanneswuerbach oh! That is a good point, no our stance is the same. Thanks for catching!

stefanpenner:
we may need to revert.

johanneswuerbach:
Seems like the latest socket.io 2 version reverted the native dependency, will give it a try once back and then either revert or upgrade further

johanneswuerbach:
Looks like there is one missing piece, pinged the socket.io maintainers and if we get a timely answer will either revert or upgrade https://github.com/socketio/socket.io/issues/3213

johanneswuerbach:
New version cut and released.

PR #1234, opened 2018-04-13
Incremental conversion of codebase to es6 classes - tap_consumer

scalvert:
Currently the codebase is largely utilizing es5 objects w/prototypical inheritance. This PR starts the conversion to es6 classes to help improve code cleanliness and readability.

Converts tap_consumer.js.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -57,26 +57,6 @@ class Config {
     }
   }
 
-  get defaults() {
-    return {
-      host: 'localhost',
-      port: 7357,
-      url() {
-        var scheme = 'http';
-        if (this.get('key') || this.get('pfx')) {
-          scheme = 'https';
-        }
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -57,6 +57,26 @@ class Config {
     }
   }
 
+  get defaults() {
```
stefanpenner: This isn't quite functionality equivalent. The accessor returns a new POJO per invocation, and disallows overrides.

```js
class A {
  get b() { return {} }
}

a = new A
> A {}
a.b = 1
> 1
a.b
> {}
```

I think, this me need to be returned to the prototype.

scalvert: Hmmmm good point. Will fix. I've been using `lebab` to convert, and i think it may have messed something up.

scalvert:
Reverted, @stefanpenner

PR #1246, opened 2018-04-24
Added support for browser_exes

Techn1x:
PR for my comment in https://github.com/testem/testem/issues/1241
Similar to https://github.com/testem/testem/pull/1244

This PR adds support for a `browser_exes` arg, so that you can supply a custom binary name for a browser without needing to specify a path (or in addition to specifying path)

This is useful for puppeteer installs, which installs the Chrome browser as `chrome` (rather than what testem expects, which is `google-chrome`)

And for good measure I also added `chrome` to the default possibleExe list anyway

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,11 +342,14 @@ This displays the current list of launchers that are available. Launchers can la
 
 Customizing Browser Paths
 -----------------------------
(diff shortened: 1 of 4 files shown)


stefanpenner:
LGTM, thanks!

released as v2.4.0 🎉

PR #1248, opened 2018-05-11
Adding configuration to set socket heartbeat timeout

step2yeung:
Adding configuration to set socket heartbeat timeout.
This gets set in browser_test_runner when the browser-login event is handled, which is when the client socket is recognized as connected.

In socketio, the `heartbeat timeout` value is used for the amount of time allowed for the both the server and client side to respond to a heartbeat request. If this timeout is exceeded, such as any action that causes the main thread to be blocked, the socket will close and a disconnect event will be emitted. This cause testem to exit with `browser timeout` error.

Due to socket disconnecting as part of tests execution, making the heartbeat timeout configurable can give the socket more time to respond to the heartbeat requests, potentially mitigating the socket disconnections.


step2yeung:
@stefanpenner to review

PR #1256, opened 2018-05-30
Server now respects the host options

### LINKED ISSUE #1255: Ember Cli test --host option is ignored by testem
Hi, I noticed that I was broadcasting my testing server to my whole network. I opened an issue in ember-cli/ember-cli and we realized that testem is not passing in host to it's server argument.

Here's how to reproduce this issue:

```
nvm use v9.3.0 
npm install -g ember-cli@latest
mkdir hello
cd hello
ember init
ember test --serve --host 127.0.0.1
lsof -i -n -P | grep 7357
```

Observed output:
```
node      13529 robertwebb   22u  IPv6 0xXXXXXXXXXXXXX     0t0  TCP *:7357 (LISTEN)
node      13529 robertwebb   30u  IPv6 0xXXXXXXXXXXXXX      0t0  TCP 127.0.0.1:7357-127.0.0.1:56246 (ESTABLISHED)
```
Expected output:
```
node      13529 robertwebb   22u  IPv6 0xXXXXXXXXXXXXX      0t0  TCP 127.0.01:7357 (LISTEN)
node      13529 robertwebb   30u  IPv6 0xXXXXXXXXXXXXX     0t0  TCP 127.0.0.1:7357->127.0.0.1:56246 (ESTABLISHED)
```

Thankfully Robert Jackson found where the issue is in testem, it's right here: https://github.com/testem/testem/blob/ed17398452743afd0a4952d343c10bf07588ea7a/lib/server/index.js#L64

It should be passing in `this.config.get('host')`

Would you like a PR for this?

robert-j-webb:
Closes #1255 

As a consequence, will now default to using localhost as opposed to 0.0.0.0


robert-j-webb:
The failure is most likely due to the EADDRINUSE error not triggering because of this line:

https://github.com/testem/testem/blob/af8620df28a58b97633957f25eb7cd3a57098013/tests/ci/ci_tests.js#L289

I think if listen is called with 127.0.0.1 it should pass the test.

stefanpenner:
> As a consequence, will now default to using localhost as opposed to 0.0.0.0

For some this may be a breaking change.

robert-j-webb:
Yes, if you set the default to 0.0.0.0 instead, it should have the same behavior as before. From a security viewpoint, it's problematic, but it would be non-breaking.

stefanpenner:
> From a security viewpoint, it's problematic

I believe if someone is genuinely concerned, they should utilize their firewall. For others, especially those who utilize VM's or commonly attach mobile devices, this ends up being a good default.

stefanpenner:
> Yes, if you set the default to 0.0.0.0 instead, 

Let's do that.

robert-j-webb:
@stefanpenner - I set the default to 0.0.0.0 - Let me know if you would like anything else! I'm happy to help :)

stefanpenner:
I think your history is funky, i would suggest you rebase master on top of yours. Rather then including master commits in your PR.

stefanpenner:
Also tests seem to fail

robert-j-webb:
Ok let me fix it

stefanpenner:
thanks for the tests!

robert-j-webb:
Thanks! Sorry for needing so much back and forth!

robert-j-webb:
I can't figure out how to fix this - the npm install command appears to be timing out, but the timeout option doesn't appear to be respected. Integration tests are fine locally, I think it's just appveyor that's too slow.

`run-integration.js` runs npm install with a timeout of 180000
```js
var TIMEOUT = 180000; // npm install is sometimes really slow...
```
But appveyor is reporting that it's timing out the integration build at just 11400ms, so it looks like it's not respecting the timeout command! I do not understand it.

stefanpenner:
> I can't figure out how to fix this - the npm install command appears to be timing out, but the timeout option doesn't appear to be respected. Integration tests are fine locally, I think it's just appveyor that's too slow.

Very strange. I'll try and find time to look.

robert-j-webb:
Thank you so much! I really loathe errors that can only be produced on test
servers.

robert-j-webb:
@stefanpenner - Any updates on this? I might be able to dig in again this weekend if you'd like. Another one of my teammates got pinged by security because they had a test server that was broadcasting to the whole network.

stefanpenner:
@robert-j-webb unfortunately, I have not had success or enough time to have success :(

Issue #1247, opened 2018-04-26
When the socket from a browser timeout, reconnection of the socket doesn't get handled by testem


step2yeung:
I am experiencing browser timeout inconsistently as part of my test execution.
I was able to observe that when this happens, the socket was closed and testem receives a disconnect event that starts the timer for the browser timeout error.
During this time, socketio will reconnect the socket via a new id. But since testem does not handle the reconnection, the timeout will expire and the browser timeout error will be thrown.

To reproduce this, I added debug logs from socket.io in both server & client side:
`DEBUG=socket.io:*` when executing test command
And on the browser, setting `localStorage.debug = '*'`

From Client side, I get logs as such:
>engine.io-client:socket socket close with reason transport close
socket.io-client:manager onclose
socket.io-client:manager cleanup
socket.io-client:manager cleanup transport close
socket.io-client:manager will wait %dms before reconnect attempt 579
socket.io-client:manager attempting reconnect
socket.io-client:manager readyState closed
socket.io-client:manager opening localhost:7357
engine.io-client:socket creating transport polling
engine.io-client:polling polling
engine.io-client:polling-xhr xhr poll
engine.io-client:polling-xhr xhr open GET http://localhost:7357/socket.io/?EIO=3&transport=polling&t=MBpRduP
engine.io-client:polling-xhr xhr data null
engine.io-client:socket setting transport polling
socket.io-client:manager connect attempt will timeout after 20000
engine.io-client:polling polling got data "96:0{\"sid\":\"0kZ0foLaS_CiaQf1AAAQ\",\"upgrades\":[\"websocket\"],\"pingInterval\":25000,\"pingTimeout\":5000}2:40"
engine.io-client:socket socket receive: type open
engine.io-client:socket socket open
socket.io-client:manager open
socket.io-client:manager cleanup 
socket.io-client:socket transport is open - connecting
socket.io-client:manager reconnect success

From the testem side, i get:

>2018-04-23T16:55:27.515Z socket.io:server incoming connection with id 0kZ0foLaS_CiaQf1AAAQ
2018-04-23T16:55:27.515Z socket.io:client connecting to namespace /
2018-04-23T16:55:27.515Z socket.io:namespace adding socket to nsp /
2018-04-23T16:55:27.515Z socket.io:socket socket connected - writing packet
2018-04-23T16:55:27.515Z socket.io:socket joining room 0kZ0foLaS_CiaQf1AAAQ
[...]

Issue #1021, opened 2016-11-19
Testem Chrome Timeouts


patocallaghan:
Hey, 

as I mention in https://github.com/testem/testem/issues/1015 we frequently get testem failures in CI when running against the Chrome (we use Testem through Ember CLI). Our test runs either fail to start (as if testem can't find the browser) or fail to finish (all tests have run and pass, but the browser window sits there and it eventually times out). It usually just tells us `testem.js has not loaded`.

As the issue is intermittent it can be hard to reproduce. I've tried to recreate it in a repo by simply [forking Travis CI's web repo](https://github.com/patocallaghan/travis-web/tree/ember-exam) (uses Ember CLI) and adding `ember-exam` to it (Ember Exam just allows you to easily use Testem's parallel feature). If running a single test run it'd probably take some time to get a timeout, but if you run `ember exam --split=9 --parallel` you should start getting some timeouts in a few runs.

I've generated some (pretty verbose) logs [here](https://gist.githubusercontent.com/patocallaghan/8cc79eeaac9a3377f44840dec15391c0/raw/1a6ab68b78217cd41ced2e52d4439b529f945cc9/gistfile1.txt).

Generally on startup I've seen the following errors:

```
not ok 323 Chrome - error
    ---
        message: >
            Error: Browser failed to connect within 30s. testem.js not loaded?
            Stderr:
             2016-11-19 00:42:01.877 Google Chrome[28575:622538] NSWindow warning: adding an unknown subview: <FullSizeContentView: 0x7fa45b8022a0>. Break on NSLog to debug.
            2016-11-19 00:42:01.878 Google Chrome[28575:622538] Call stack:
            (
                "+callStackSymbols disabled for performance reasons"
            )


        Log: |
            { type: 'error',
              text: 'Error: Browser failed to connect within 30s. testem.js not loaded?' }
            { type: 'error',
              text: '2016-11-19 00:42:01.877 Google Chrome[28575:622538] NSWindow warning: adding an unknown subview: <FullSizeContentView: 0x7fa45b8022a0>. Break on NSLog to debug.\n2016-11-19 00:42:01.878 Google Chrome[28575:622538] Call stack:\n(\n    "+callStackSymbols disabled for performance reasons"\n)\n' }
    ...
```

On timeouts at the end I've seen

```
  engine:ws writing "3" +1ms
  engine:polling closing +649ms
  engine:polling transport not writable - buffering orderly close +0ms
  socket.io:client client close with reason ping timeout +0ms
[...]

PR #1257, opened 2018-06-01
Generate unique id in launcher-factory, app.js to use launch-factory

step2yeung:
Problem: Launcher currently generates the id on every browser-login, but nothing keeps track of the id that is generated, and it could cause a collision. When this happens, I have experienced a browser timeout error.

Solution: In launcher-factory, keep track of the id that are used. And if it is already in use, generate a new one.
Change app.js to use launcher-factory so the above is used.

This should also help with #1021 #1247.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -14,7 +14,7 @@ const TapProcessTestRunner = require('./runners/tap_process_test_runner');
 const HookRunner = require('./runners/hook_runner');
 const cleanExit = require('./clean_exit');
 const FileWatcher = require('./file_watcher');
-const Launcher = require('./launcher');
+const LauncherFactory = require('./launcher-factory');
 
 const RunTimeout = require('./utils/run-timeout');
 const Reporter = require('./utils/reporter');
@@ -238,10 +238,10 @@ module.exports = class App extends EventEmitter {
     });
 
     if (!browser) {
-      let launcher = new Launcher(browserName, {
+      let launcher = new LauncherFactory(browserName, {
         id: id,
         protocol: 'browser'
-      }, this.config);
+      }, this.config).create();
       browser = new BrowserTestRunner(launcher, this.reporter, this.runnerIndex++, this.config);
       this.addRunner(browser);
     }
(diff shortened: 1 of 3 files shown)


stefanpenner:
@dcombslinkedin idea of a `monotonically-increasing number` sounds like a good solution.

If we are concerned about multiple concurrent testem runs colliding, let me recommend 1 random seed per testem + a monotonically increasing number.

```js
const TESTEM_SESSION = 'TESTEM_' + Date.now() + '-';
let _maxId = 0;

function getUniqueId() {
   return `${TESTEM_SESSION}-${_maxId++}`;
}
```

This should:

* ensure no pollution between testem modules
* provide an easy to ready incrementing id per instance

stefanpenner:
@step2yeung  / @dcombslinkedin good sleuthing!

step2yeung:
@stefanpenner @dcombslinkedin I updated the code to use process.hrtime() as the uniqueId. The problem with using `Date.now()` is that its possible for each browser to start fast enough for it to not be unique.
As for @stefanpenner's suggestion with using a string with monotonically-increasing number for the id, the code as is currently doesnt support using a string id, and making those changes may need more digging into.

dcombslinkedin:
@step2yeung , the launcher-factory.js file doesn't yet have the use of process.hrtime(), it still shows the use of Math.random(). Did you check in and push?

stefanpenner:
> @stefanpenner @dcombslinkedin I updated the code to use process.hrtime() as the uniqueId. The problem with using Date.now() is that its possible for each browser to start fast enough for it to not be unique.

@step2yeung good call!

step2yeung:
@dcombslinkedin good call! Pushed correctly now.

stefanpenner:
@step2yeung LGTM, mind adding a quick unit test? Then I can merge and release.

step2yeung:
@stefanpenner added launcher-factory tests.

stefanpenner:
LGTM, small tweak: https://github.com/testem/testem/pull/1263

stefanpenner:
released as v2.7.0 🎉

PR #1262, opened 2018-06-04
fix(teamcity_reporter): nest tests by prefix

### LINKED ISSUE #1187: Possibly nest teamcity suites by launcher
Hi,
the teamcity spec allows to [nest test suites](https://confluence.jetbrains.com/display/TCD65/Build+Script+Interaction+with+TeamCity#BuildScriptInteractionwithTeamCity-ReportingTests):

```
##teamcity[testSuiteStarted name='suite.name']
##teamcity[testSuiteStarted name='nested.suite']
##teamcity[testStarted name='package_or_namespace.ClassName.TestName']
##teamcity[testFailed name='package_or_namespace.ClassName.TestName' message='The number should be 20000' details='junit.framework.AssertionFailedError: expected:<20000> but was:<10000>|n|r    at junit.framework.Assert.fail(Assert.java:47)|n|r    at junit.framework.Assert.failNotEquals(Assert.java:280)|n|r...']
##teamcity[testFinished name='package_or_namespace.ClassName.TestName']
##teamcity[testSuiteFinished name='nested.suite']
##teamcity[testSuiteFinished name='suite.name']
```

The current implementation only emits a flat list of test results. 
I'm unsure if we could nest the output because this requires the reporter to wait for all tests inside a launcher to be finished.

Nested output allows an easier navigation through test results. See for example:

Which is nested by grouping the results by their first `/(.*) - /` match.

makepanic:
fixes #1187

### DIFF
--- a/lib/reporters/teamcity_reporter.js
+++ b/lib/reporters/teamcity_reporter.js
@@ -11,11 +11,13 @@ class TeamcityReporter {
     this.skipped = 0;
     this.startTime = new Date();
     this.endTime = null;
+    this.prefixedResults = {};
   }
 
   report(prefix, data) {
-    this.out.write('##teamcity[testStarted name=\'' + this._namify(prefix, data) + '\']\n');
-    this._display(prefix, data);
+    var _prefix = prefix ? prefix : 'testem.suite';
+    this.prefixedResults[_prefix] = this.prefixedResults[_prefix] || [];
+    this.prefixedResults[_prefix].push(data);
     this.total++;
     if (data.skipped) {
       this.skipped++;
@@ -29,30 +31,41 @@ class TeamcityReporter {
       return;
     }
     this.endTime = new Date();
-    this.out.write('\n\n');
-    this.out.write('##teamcity[testSuiteFinished name=\'testem.suite\' duration=\'' + this._duration() + '\']\n');
+
+    for (var prefix in this.prefixedResults) {
+      if (this.prefixedResults.hasOwnProperty(prefix)) {
+        this.out.write('##teamcity[testSuiteStarted name=\''+prefix+'\']\n');
+        this.prefixedResults[prefix].forEach(this._display.bind(this));
+        this.out.write('##teamcity[testSuiteFinished name=\''+prefix+'\']\n');
+      }
+    }
+
     this.out.write('\n\n');
   }
 
-  _display(prefix, result) {
+  _display(result) {
     if (this.silent) {
       return;
     }
+
+    var name = this._namify(result);
+
+    this.out.write('##teamcity[testStarted name=\'' + name + '\']\n');
+
     if (result.skipped) {
-      this.out.write('##teamcity[testIgnored name=\'' + this._namify(prefix, result) + '\' message=\'pending\']\n');
+      this.out.write('##teamcity[testIgnored name=\'' + name + '\' message=\'pending\']\n');
     } else if (!result.passed) {
       var message = (result.error && result.error.message) || '';
       var stack = (result.error && result.error.stack) || '';
-      this.out.write('##teamcity[testFailed name=\'' + this._namify(prefix, result) + '\' message=\'' + escape(message) + '\' details=\'' + escape(stack) + '\']\n');
-    }
-    this.out.write('##teamcity[testFinished name=\'' + this._namify(prefix, result) + '\'' + this._runDurationAttribute(result) + ']\n');
 
+      this.out.write('##teamcity[testFailed name=\'' + name + '\' message=\'' + escape(message) + '\' details=\'' + escape(stack) + '\']\n');
+    }
+
+    this.out.write('##teamcity[testFinished name=\'' + name + '\'' + this._runDurationAttribute(result) + ']\n');
   }
 
-  _namify(prefix, result) {
-    var line = (prefix ? (prefix + ' - ') : '') +
-      result.name.trim();
-    return escape(line);
+  _namify(result) {
+    return escape(result.name.trim());
   }
 
   _duration() {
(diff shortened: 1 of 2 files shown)


stefanpenner:
@turbo87 this appears related to: https://github.com/Turbo87/intellij-emberjs/pull/166 mind confirming this is good?

makepanic:
To confirm that the reporter changes still work, I've loaded it up in ember-intellij with the patched test run configuration:

stefanpenner:
@makepanic awesome. That looks great. Once I get @turbo87's +1, i'll merge and ship. :)

tests/ci/reporter_tests.js
```diff
@@ -504,18 +504,61 @@ describe('test reporters', function() {
...
-      assert.match(output, /##teamcity\[testIgnored name='phantomjs - it skips stuff' message='pending']/);
-      assert.match(output, /##teamcity\[testFinished name='phantomjs - it skips stuff']/);
-      assert.match(output, /##teamcity\[testStarted name='phantomjs - it handles failures']/);
-      assert.match(output, /##teamcity\[testFailed name='phantomjs - it handles failures' message='foo' details='bar']/);
-      assert.match(output, /##teamcity\[testFinished name='phantomjs - it handles failures']/);
-      assert.match(output, /##teamcity\[testStarted name='phantomjs - it handles undefined errors']/);
-      assert.match(output, /##teamcity\[testFailed name='phantomjs - it handles undefined errors' message='' details='']/);
-      assert.match(output, /##teamcity\[testFinished name='phantomjs - it handles undefined errors' duration='42']/);
+      assert.match(output, /##teamcity\[testSuiteStarted name='phantomjs']/);
+      assert.match(output, /##teamcity\[testSuiteFinished name='phantomjs']/);
```
Turbo87: I understand that this is currently not testing the order in which these messages are reported, but I guess that might be a good idea to make sure that `testSuiteFinished` is only printed *after* all the tests have finished 🤔

lib/reporters/teamcity_reporter.js
```diff
@@ -29,30 +31,41 @@ class TeamcityReporter {
...
     }
     this.endTime = new Date();
-    this.out.write('\n\n');
-    this.out.write('##teamcity[testSuiteFinished name=\'testem.suite\' duration=\'' + this._duration() + '\']\n');
+
+    for (var prefix in this.prefixedResults) {
+      if (this.prefixedResults.hasOwnProperty(prefix)) {
+        this.out.write('##teamcity[testSuiteStarted name=\''+prefix+'\']\n');
+        this.prefixedResults[prefix].forEach(this._display.bind(this));
+        this.out.write('##teamcity[testSuiteFinished name=\''+prefix+'\']\n');
```
Turbo87: is it intended that `this._duration()` was dropped from the `testSuiteFinished` event?

makepanic: I think previously it was an error as the [jetbrains docs](https://confluence.jetbrains.com/display/TCD65/Build+Script+Interaction+with+TeamCity#BuildScriptInteractionwithTeamCity-ReportingTests) mention `duration` only for *Test start/stop messages*

tests/ci/reporter_tests.js
makepanic: I think what you're asking about is tested in the next test.

Turbo87:
@makepanic if I understand this patch correctly then previously it would write out test results as they are reported. with this change it would only write out the results at the end of the whole test suite. does that have any negative implications for tools that want to report the test results "live"?

makepanic:
> with this change it would only write out the results at the end of the whole test suite. does that have any negative implications for tools that want to report the test results "live"?

Right that could be an issue.
With this PR the behavior changes from writing the result when it arrives to buffering until the tests are finished, similar to the [xunit reporter](https://github.com/testem/testem/blob/master/lib/reporters/xunit_reporter.js#L62).

stefanpenner:
> With this PR the behavior changes from writing the result when it arrives to buffering until the tests are finished, similar to the xunit reporter.

That seems a tad unfortunate for those with long/large test suites

makepanic:
I guess one could add a new reporter parallel to the old teamcity reporter.

Dunno if it's a good idea to ship it in testem or as an extra package.

Turbo87:
@makepanic can you check if IntelliJ merges test suites with the same name? then we could open a new test suite per test, and let IntelliJ worry about merging them up so that tests are grouped per suite 🤔

makepanic:
> check if IntelliJ merges test suites with the same name

Great idea.
I've tested it but sadly they generate tons of nested suites containing only one test.
I've also tried to use [`flowId`](https://confluence.jetbrains.com/display/TCD10/Build+Script+Interaction+with+TeamCity#BuildScriptInteractionwithTeamCity-MessageFlowId) and emitting suiteStarted/Finished with the defined [test name format](https://confluence.jetbrains.com/display/TCD10/Build+Script+Interaction+with+TeamCity#BuildScriptInteractionwithTeamCity-Interpretingtestnames) but that also didn't work :<

I guess the teamcity format requires input to be xml'ish.

Dunno if the additional overhead of maintaining a new test reporter, additional configuration or introducing a breaking change is worth it tbh.

Turbo87:
hmm okay. guess then there is not much we can do about this unfortunately. I'm fine either way, the implementation looks good to me. 

@makepanic I guess it's up to you. If you think this is the right way forward than I'm 👍on merging this.

stefanpenner:
@makepanic in theory, we could make this opt-in? (if there is concern about backwards compat)

stefanpenner:
@makepanic with opt-in, i see not reason not to merge + release immediately once complete. Just be sure to add the appropriate docs :)

PR #1268, opened 2018-06-07
Moving set heartbeat timeout from browser_test_runner to server/index

step2yeung:
The original place where heartbeat_timeout is set (in browser_test_runner.js) is too late to set the heartbeat_timeout for the first browser instance. 

Problem: The first browser is spawned with the default timeout value. BrowserLogin event is emitted and handled, then testem tell the server to set the subsequent browsers to use the newly configured heartbeat timeout value. 

Fix: move the configuration code earlier, right after the testem server socket is initiated.

### DIFF
--- a/lib/runners/browser_test_runner.js
+++ b/lib/runners/browser_test_runner.js
@@ -101,9 +101,6 @@ module.exports = class BrowserTestRunner {
 
     this.onStart.call(this);
 
-    let socketHeartbeatTimeout = this.launcher.config.get('socket_heartbeat_timeout');
-    socket.server.set('heartbeat timeout', socketHeartbeatTimeout * 1000);
-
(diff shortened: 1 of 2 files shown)


stefanpenner:
@step2yeung mind adding a unit test?

tests/server_tests.js
```diff
@@ -174,6 +175,11 @@ describe('Server', function() {
       });
     });
 
+    it('sets heartbeat_timeout on socket.io server', function(done){
+      expect(server.io.eio['pingTimeout']).to.eq(10000);
```
stefanpenner: `done` is redundant here, if you can remove i'll merge and release

stefanpenner:
released as v2.7.2 🎉

PR #1274, opened 2018-06-20
adding better error handling when test-suite timeout occurs

dcombslinkedin:
Recently we were running our large (many thousands of tests) test suite on an Ember app, and intermittently got a 'Browser exited unexpectedly' error. We finally tracked it down to we set the testem 'timeout' config value to a number not high enough to get through all the tests the browser was to run (we're also partitioning tests across browsers, which confused things as more tests were added over time). The problem was that testem did not put anything in the log when it decided to shut down the browser due to expiring the timeout value. The browser onProcessExit handler then put out its normal "browser exited unexpectedly" message, which we thought for a time meant "the browser crashed" (which is wrong).

This fix is to add a flag when testem itself kills the browser, and some handling in onProcessExit to change the exit message when the flag is true, to provide context for why the process exited early. This would have saved us a week of debugging. ;-)

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -150,7 +150,14 @@ module.exports = class App extends EventEmitter {
 
     return Bluebird.using(this.runHook('before_tests'), () => {
       return Bluebird.using(RunTimeout.with(this.config.get('timeout')), timeout => {
-        timeout.on('timeout', () => this.killRunners());
+        timeout.on('timeout', () => {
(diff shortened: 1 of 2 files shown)


dcombslinkedin:
Thank you, @stefanpenner ! Much appreciated.

stefanpenner:
release as 2.8.2

PR #1276, opened 2018-06-21
Throwing error for window.alert and window.confirm

### LINKED ISSUE #1275: explore stubbing out `alert` `confirm` etc
If these are invoked, testem cannot recover. We should consider just replacing them in `testem.js` with functions that error with a useful error message.

step2yeung:
This will notify developers that window.alert and window.confirm will cause a browser disconnect failures.

Fixes #1275


stefanpenner:
Awesome, I was just about to implement. What a pleasant surprise to see it already done!

PR #1281, opened 2018-07-19
Add better reporting for qunit negative assertions

jamescdavis:
For negative assertions, qunit sets `negative: true`, e.g. [notEqual](https://github.com/qunitjs/qunit/blob/a577183e8d4f975d822e3021cbba552c5ed77549/src/assert.js#L189).

This PR passes along the `negative` key to reporters and prepends `NOT` to the output for `expected` when the `negative` key is truthy in the *dot* and *teamcity* reporters as well as the dev mode reporter (the *tap* and *xunit* reporters do not output `expected`).

This brings testem in parity with qunit's own [html reporter](https://github.com/qunitjs/qunit/blob/a577183e8d4f975d822e3021cbba552c5ed77549/reporter/html.js#L773).

### DIFF
--- a/lib/reporters/dev/split_log_panel.js
+++ b/lib/reporters/dev/split_log_panel.js
@@ -29,7 +29,7 @@ function failureDisplay(item) {
   }
 
   if (item.expected) {
-    extra.push(' expected ' + item.expected);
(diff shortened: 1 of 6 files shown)


johanneswuerbach:
Thanks, released as `v2.9.0`

PR #1283, opened 2018-08-02
[FEAT BUGFIX] add test time and duration to TapReporter, fix double-exit check, print timeout error on exit if required

runspired:
Previous testem output:

```
ok 1 phantomjs - My Awesome Test Scenario
```

New Output

```
[12:31:30] ok 1 phantomjs - [50 ms] - My Awesome Test Scenario
```

Previous global timeout error message:

```
Browser exited on request from test driver
```

New global timeout error message

```
Test suite execution has timed out (config.timeout = 300 seconds). Terminated all test runners.
```

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -33,11 +33,14 @@ module.exports = class App extends EventEmitter {
     this.runnerIndex = 0;
     this.runners = [];
     this.timeoutID = undefined;
+    this.testSuiteTimedOut = null;
+    this.testSuiteTimedOut = false;
 
     this.reportFileName = this.config.get('report_file');
 
+    let alreadyExit = false;
+
     this.cleanExit = err => {
-      let alreadyExit = false;
       if (!alreadyExit) {
         alreadyExit = true;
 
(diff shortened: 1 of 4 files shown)


runspired:
cc @dcombslinkedin @stefanpenner @rwjblue

lib/app.js
```diff
@@ -47,6 +50,11 @@ module.exports = class App extends EventEmitter {
           err = null;
         }
 
+        if (this.testSuiteTimedOut === true) {
+          let timeoutSeconds = this.testSuiteTimeout.timeout;
+          err = new Error(`Test suite execution has timed out (config.timeout = ${timeoutSeconds} seconds). Terminated all test runners.`);
+        }
```
runspired: I think I need to also force `exitCode` to `1` here

stefanpenner:
@runspired have you confirmed this displays in the right place, in an actionable way? Specifically, would this have made your debugging session earlier this much better?

runspired:
screen shot 2018-08-02 at 2 12 01 pm

@stefanpenner there may be more we could do but this definitely would have been a nice start :)

stefanpenner:
released as v2.9.1 :tada:

kategengler:
The prefix of the time means this output no longer follows the [TAP spec](https://testanything.org/tap-specification.html)

> ok or not ok
This tells whether the test point passed or failed. **It must be at the beginning of the line**. /^not ok/ indicates a failed test point. /^ok/ is a successful test point. This is the only mandatory part of the line. Note that unlike the Directives below, ok and not ok are case-sensitive. (emphasis added)

I suspect this will break various tools that use TAP output.

runspired:
#1284 exists to rectify this

kategengler:
@runspired I think the time in that location is fine (it would count as part of the test description, but so does `phantomjs`).

Issue #1286, opened 2018-08-15
Forward slashes in the cwd option cause "safe directory" check to fail erroneously.


the-t-in-rtf:
If you specify a value for the `cwd` option that contains forward slashes, such as `c:/path/to/fixtures` and then run Testem from a directory higher than that path, the tests will fail with 403 errors, because of [this "safe dirs" check](https://github.com/testem/testem/blob/master/lib/server/index.js#L345)

The check will fail for a `cwd` value that contains forward slashes because one half of the comparison is run through `path.resolve` (which standardises the slashes), and one half (`config.cwd()`) isn't:

```
var filePath = path.resolve(config.resolvePath(uri));
    var ext = path.extname(filePath);
    var isPathPermitted = filePath.indexOf(config.cwd()) !== -1;
```
I will submit a pull that wraps the call to `config.cwd()` in a similar call to `path.resolve`, I have verified that this resolves the problem in local testing.

PR #1287, opened 2018-08-15
GH-1286: Added path resolve of cwd to "safe dir" check.

the-t-in-rtf:
See [the associated writeup](https://github.com/testem/testem/issues/1286) for details.


johanneswuerbach:
Could you add a simple test for this? So we don't regress,

the-t-in-rtf:
Sure, will do (probably tomorrow morning, though).

johanneswuerbach:
Thanks, published as `v2.9.3` :-)

the-t-in-rtf:
Awesome!  I wrote the new tests to cover the unsafe dir handling, which didn't seem to be tested already.  I verified manually in Windows that reverting the change results in the request for an allowed piece of content (within the cwd) failing.

Issue #1075, opened 2017-02-21
`setRawMode` error on exit


NicolasPelletier:
When I am using TestEm in development mode ( `testem` ), I get the following error when I press `q` to quit:
```
[Press ENTER to run tests; q to quit; p to pause]                                                                                                                                                                                                                                      q
/Users/pelletn/<MyProject>/node_modules/bluebird/js/release/using.js:12
        setTimeout(function(){throw e;}, 0);
                              ^

TypeError: Cannot read property 'setRawMode' of null
    at ReadStream.setRawMode (tty.js:38:15)
    at module.exports.View.extend.setRawMode (/Users/pelletn/<MyProject>/node_modules/testem/lib/reporters/dev/index.js:333:21)
    at module.exports.View.extend.cleanup (/Users/pelletn/<MyProject>/node_modules/testem/lib/reporters/dev/index.js:398:10)
    at module.exports.View.extend.finish (/Users/pelletn/<MyProject>/node_modules/testem/lib/reporters/dev/index.js:388:10)
    at /Users/pelletn/<MyProject>/node_modules/testem/lib/utils/reporter.js:101:22
    at Array.forEach (native)
    at Reporter.finish (/Users/pelletn/<MyProject>/node_modules/testem/lib/utils/reporter.js:99:20)
    at Reporter.close (/Users/pelletn/<MyProject>/node_modules/testem/lib/utils/reporter.js:77:8)
    at Reporter.<anonymous> (/Users/pelletn/<MyProject>/node_modules/testem/lib/utils/reporter.js:72:21)
    at FunctionDisposer.doDispose (/Users/pelletn/<MyProject>/node_modules/bluebird/js/release/using.js:98:19)
    at FunctionDisposer.Disposer.tryDispose (/Users/pelletn/<MyProject>/node_modules/bluebird/js/release/using.js:78:20)
    at iterator (/Users/pelletn/<MyProject>/node_modules/bluebird/js/release/using.js:36:53)
    at dispose (/Users/pelletn/<MyProject>/node_modules/bluebird/js/release/using.js:48:9)
    at /Users/pelletn/<MyProject>/node_modules/bluebird/js/release/using.js:194:20
    at PassThroughHandlerContext.finallyHandler (/Users/pelletn/<MyProject>/node_modules/bluebird/js/release/finally.js:55:23)
    at PassThroughHandlerContext.tryCatcher (/Users/pelletn/<MyProject>/node_modules/bluebird/js/release/util.js:16:23)
```

[...]

PR #1289, opened 2018-08-21
Default 2nd reporter type in dev mode to tap or use dev_mode_file_reporter conf

ventuno:
Fixes https://github.com/testem/testem/issues/1075.

See https://github.com/testem/testem/issues/1075#issuecomment-412674637 for a detailed description of the issue.

@johanneswuerbach I'm still not 100% sure this is the best solution. I like the idea that we are logging to a file, but assuming a `tap` reporter will always work fine seems wrong to me. What if the `report_file` is an `XML` file?

I think if we really want to have two reporters we might as well have a config property like `dev_mode_file_reporter` that the user can set to the preferred reporter type. Let me know what you think, otherwise this PR does just what we discussed on #1075.

### DIFF
--- a/docs/config_file.md
+++ b/docs/config_file.md
@@ -75,6 +75,7 @@ Chrome, Chrome Canary, Chromium, Firefox, IE, Opera, PhantomJS, Safari, Safari T
     config_dir:                  [Path]    directory to use as root for resolving configs, if different than cwd
     css_files:                   [Array]   string or array of additional stylesheets to include
     cwd:                         [Path]    directory to use as root
+    dev_mode_file_reporter:      [String]  in dev mode the default reporter is 'dev' for the standard output. It is possible to specify a custom reporter which will report to report_file ('tap' reporter is used otherwise)
     disable_watching:            [Boolean] disable any file watching
     fail_on_zero_tests:          [Boolean] whether process should exit with error status when no tests found
     firefox_user_js:             [String]  path to firefox custom user.js file to be used
--- a/lib/utils/reporter.js
+++ b/lib/utils/reporter.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const Bluebird = require('bluebird');
+const log = require('npmlog');
 
 const reporters = require('../reporters');
 const isa = require('../isa');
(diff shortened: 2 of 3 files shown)


johanneswuerbach:
I think in the long term creating multiple exports should be exposed in the config (the code should already support that), but I think that should be fine for now.

ventuno:
Thanks @johanneswuerbach, I'm not sure what you mean by "creating multiple exports should be exposed in the config", but if you think a new config `dev_mode_file_reporter` will be fine, I'll add that.

johanneswuerbach:
Yes, that also sounds fine for now.

> multiple exports should be exposed in the config

I was think of exposing the ability to specify an array of reports, which you could use like this:

```js
reporters = [{
   type: 'tap',
   output: 'stdout',
}, {
   type: 'junit',
   output: 'file',
   path: 'test_results.xml'
}]
```

This should be already supported within testem itself, but you can configure multiple reporters yet using the config / command line.

ventuno:
Thanks a lot @johanneswuerbach, I just added support for the `dev_mode_file_reporter` configuration option which lets the user define a custom `reporter` which will write to `report_file` in `dev` mode.
I also updated [config_file.md](https://github.com/testem/testem/blob/master/docs/config_file.md) with the documentation for the new option.

One last thing could be: do we really want to default to the `tap` reporter if none is specified or should we simply ignore it and maybe log a warning and invite the user to use the `dev_mode_file_reporter` option instead?

ventuno:
@johanneswuerbach any chance you could take a look at this?

johanneswuerbach:
Sorry, the warning sounds like a good idea.

ventuno:
Thanks @johanneswuerbach, I added the log as well. Would you mind taking reviewing this and see if this can be merged?

ventuno:
Thanks @johanneswuerbach any idea when the next release is going to happen?

johanneswuerbach:
Released as `v2.10.0`, thank you 👏 👍

ventuno:
Thanks @johanneswuerbach finally managed to test the latest version: works great!

PR #1295, opened 2018-09-05
Sending qunit testId in test-result event and exposing originalResultObject to reporter.

step2yeung:
This will allow custom reporters to report the testId in qunit, which can be used to rerun tests.


stefanpenner:
released as v2.11.0 🎉

Issue #1280, opened 2018-07-14
Why do we kill the cache on every request?


step2yeung:
I'm curious is there a reason why the testem server kill the cache on every request?
https://github.com/testem/testem/blob/master/lib/server/index.js#L305

stefanpenner:
I suspect to avoid stale reads. What are your thoughts?

step2yeung:
If i understand the use of ETag header correctly: it is a validation token that can be used check if the resource requested has been modified since the last request.
So we can send an Etag header as part of each response to avoid stale reads?

stefanpenner:
@step2yeung yup, if we use etags correctly it would be appropriate.

PR #1297, opened 2018-09-14
Use etags on assets request header to leverage browser cache

step2yeung:
handles #1280

### DIFF
--- a/lib/server/index.js
+++ b/lib/server/index.js
@@ -145,6 +145,7 @@ class Server extends EventEmitter {
   configureExpress(app) {
     app.engine('mustache', Mustache);
     app.set('view options', {layout: false});
+    app.set('etag', 'strong');
     app.use((req, res, next) => {
       if (this.ieCompatMode) {
         res.setHeader('X-UA-Compatible', `IE=${this.ieCompatMode}`);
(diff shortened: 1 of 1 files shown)


step2yeung:
Tested manually using an ember app: 
```
ember exam --server
first load of assets should be 200 response
refresh test page
consecutive loads of assets should be 304 response
change a file
reload of the change file should be 200 response
other files should be 304 response
```

stefanpenner:
@step2yeung testing this locally, I believe it requires an additional change:

```
diff --git a/lib/server/index.js b/lib/server/index.js
index 49eea8a..0341ff1 100644
--- a/lib/server/index.js
+++ b/lib/server/index.js
@@ -235,8 +235,7 @@ class Server extends EventEmitter {
   }

   renderDefaultTestPage(req, res) {
-    res.header('Cache-Control', 'No-cache');
-    res.header('Pragma', 'No-cache');
+    app.set('etag', 'strong');
```

stefanpenner:
@step2yeung with my above change, everything does seem to work correct. If you can make that correction, tidy up the commit history, I can merge and release this afternoon :)

stefanpenner:
Actually, more specifically. It seems like we should:

* remove all per request caching
* add ` app.set('etag', 'strong');` where the express app is created

stefanpenner:
I'm not sure what is up, https://github.com/testem/testem/pull/1297#issuecomment-422162984 appears to be invalid this issue just stopped happening?

stefanpenner:
released as v2.12.0 🎉

PR #1303, opened 2018-10-08
Use tmp to create browserTmpDir so it can get cleaned up when test run completes

step2yeung:
This cleans up tmp directory that testem uses during the test execution.


stefanpenner:
released as v2.13.0 🎉

stefanpenner:
We need to find some time to fix this brittle tests/CI

PR #1307, opened 2018-10-19
register events defined in additionalBrowserSocketEvents and send events to browser side

choheekim:
The changes in this PR enables communication between browser and server for test load balancing and is part of a larger set of changes ember-exam. Here’s a link to the RFC: [Test load balancing.
](https://github.com/ember-cli/ember-exam/issues/134)

1. Binds events defined in an `additionalBrowserSocketEvents` option in testem config. 
- The `additionalBrowserSocketEvents` option is an object which maps event names to event handlers. Currently, there is only one way event communication from testem to browser; this opens communication via events originating from the browser to the testem browser instance. An example:

```js
const events = config.additionalBrowserSocketEvents || {};

events['event-name'] = function () {
   return 'foo';
}
```

* Refer this [link](https://github.com/ember-cli/ember-exam/pull/136/files#diff-87b833f28b68e5193e797dd72f03c7a8R381) for how this is used in load balancing work to handle requests for test modules via events from the browser.

2. Testem_client handles the socket communication between testem’s browser instance and the browser running the test. initSocket() whitelists the socket events the browser will handle. Since we are now allowing custom events to be handled on both sides, we need to open up the socket event handling on the browser to handle all requests.

3. `removeEventCallbacks` removes any duplicate handler functions for an event in an event handler queue in order to have unique handler functions per events.  
Example of usage - https://github.com/ember-cli/ember-exam/pull/136/files#diff-babb5d4fe8ce7520c798eaae9f5393a0R60:
The `module-queue-complete` is attached everytime moduleDone() is executed but the event is emitted once a quene is empty.  Then, the event handler queue for the `module-queue-complete`  event will contain multiple number of unhandled registered callbacks. To prevent it, `removeDuplicateCallbacks` is executed to get rid of any duplicate callbacks for an event.
```js
qunit.moduleDone(function() {
    return new Promise(function(resolve) {
      const resolveCallbackFn = function resolveCallback() {
        resolve();
      }
      Testem.once('next-module-response', function getTestModule(moduleName) {
        if (moduleName !== undefined) {
          loadIndividualModule(moduleName);
        }
        // `removeCallback` removes if the event queue has `module-queue-complete` event with a same callback.
        Testem.removeEventCallbacks('module-queue-complete', resolveCallbackFn);
        resolveCallbackFn();
      });
      Testem.on('module-queue-complete', resolveCallbackFn);
      Testem.emit('get-next-module');
    });
  });
}
```

### DIFF
--- a/docs/config_file.md
+++ b/docs/config_file.md
@@ -66,6 +66,7 @@ Chrome, Chrome Canary, Chromium, Firefox, IE, Opera, PhantomJS, Safari, Safari T
 
 ### Config-level options:
 
+    additional_browser_socket_events    [Object]  an object containing keys corresponding to event names that point to handler functions, which are to be added to the browser socket
     browser_disconnect_timeout   [Number]  timeout to error after disconnect in seconds (10s)
     browser_start_timeout        [Number]  timeout to error after browser start in seconds (30s)
     browser_paths:               [Object]  hash of browsers (keys) to an string of their binary paths (values)
--- a/docs/config_file.md
+++ b/docs/config_file.md
@@ -66,7 +66,7 @@ Chrome, Chrome Canary, Chromium, Firefox, IE, Opera, PhantomJS, Safari, Safari T
 
 ### Config-level options:
 
-    additional_browser_socket_events    [Object]  an object containing keys corresponding to event names that point to handler functions, which are to be added to the browser socket
+    custom_browser_socket_events    [Object]  an object containing keys corresponding to event names that point to handler functions, which are to be added to the browser socket
     browser_disconnect_timeout   [Number]  timeout to error after disconnect in seconds (10s)
     browser_start_timeout        [Number]  timeout to error after browser start in seconds (30s)
     browser_paths:               [Object]  hash of browsers (keys) to an string of their binary paths (values)
--- a/public/testem/testem_client.js
+++ b/public/testem/testem_client.js
@@ -175,6 +175,7 @@ var Testem = {
         handler.apply(this, argsWithoutFirst);
       }
     }
+
     this.emitMessage.apply(this, arguments);
   },
   on: function(evt, callback) {
@@ -266,6 +267,9 @@ var Testem = {
         case 'stop-run':
           self.emit('after-tests-complete');
           break;
+        default:
+          self.emit(type, message.data);
+          break;
       }
     });
   },
@@ -285,6 +289,18 @@ var Testem = {
     }
     // stringify for clients that only can handle string postMessages (IE <= 10)
     return JSON.stringify(message);
+  },
+  removeEventCallbacks: function(evt, callback) {
+    var handlers = this.evtHandlers[evt];
+    var removeIdx = [];
+    for (var i = 0; i < handlers.length; i++) {
+      if (handlers[i] === callback) {
+        removeIdx.push(i);
+      }
+    }
+    for (var j = 0; j < removeIdx.length; j++) {
+      handlers.splice(j, 1);
+    }
   },
   runAfterTests: function() {
     if (Testem.afterTestsQueue.length) {
--- a/public/testem/testem_client.js
+++ b/public/testem/testem_client.js
@@ -268,7 +268,7 @@ var Testem = {
           self.emit('after-tests-complete');
           break;
         default:
-          if (type.indexOf('testem:') == 0) {
+          if (type.indexOf('testem:') === 0) {
             self.emit(type, message.data);
           }
           break;
--- a/public/testem/testem_client.js
+++ b/public/testem/testem_client.js
@@ -295,6 +295,9 @@ var Testem = {
   removeEventCallbacks: function(evt, callback) {
     var handlers = this.evtHandlers[evt];
     var removeIdx = [];
+    if (typeof handlers === "undefined") {
+      return;
+    }
     for (var i = 0; i < handlers.length; i++) {
       if (handlers[i] === callback) {
         removeIdx.push(i);
--- a/public/testem/testem_client.js
+++ b/public/testem/testem_client.js
@@ -268,7 +268,9 @@ var Testem = {
           self.emit('after-tests-complete');
           break;
         default:
-          self.emit(type, message.data);
+          if (type.indexOf('testem:') == 0) {
+            self.emit(type, message.data);
+          }
           break;
       }
     });
--- a/public/testem/testem_connection.js
+++ b/public/testem/testem_connection.js
@@ -190,7 +190,7 @@ function initSocket(id) {
     sendMessageToParent('stop-run');
   });
   socket.on('*', function(event) {
-    if (event.data) {
+    if (event.data && event.data[0].indexOf('testem:') == 0 ) {
       var eventName = event.data[0];
       var eventData = event.data[1];
       sendMessageToParent(eventName, eventData);
--- a/public/testem/testem_connection.js
+++ b/public/testem/testem_connection.js
@@ -1,4 +1,5 @@
-/* globals document, parent, window, io, navigator */
+/* globals document, parent, io, window, navigator */
+/* globals module */
 'use strict';
 
 var socket;
@@ -37,11 +38,13 @@ function sendMessageToParent(type, data) {
   parent.postMessage(message, '*');
 }
 
-var addListener = window.addEventListener ?
-  function(obj, evt, cb) { obj.addEventListener(evt, cb, false); } :
-  function(obj, evt, cb) { obj.attachEvent('on' + evt, cb); };
-
-addListener(window, 'message', handleMessage);
+var addListener;
+if (typeof window !== 'undefined') {
+  addListener = window.addEventListener ?
+    function(obj, evt, cb) { obj.addEventListener(evt, cb, false); } :
+    function(obj, evt, cb) { obj.attachEvent('on' + evt, cb); };
+    addListener(window, 'message', handleMessage);
+}
 
 var messageListeners = {};
 function handleMessage(event) {
@@ -190,7 +193,7 @@ function initSocket(id) {
     sendMessageToParent('stop-run');
   });
   socket.on('*', function(event) {
-    if (event.data && event.data[0].indexOf('testem:') == 0 ) {
+    if (event.data && event.data[0].indexOf('testem:') === 0 ) {
       var eventName = event.data[0];
       var eventData = event.data[1];
       sendMessageToParent(eventName, eventData);
@@ -198,4 +201,12 @@ function initSocket(id) {
   });
 }
 
-init();
+// We should only call init() if it ran in browser.
+if (typeof window !== 'undefined') {
+  init();
+}
+
+// Exporting this as a module so that it can be unit tested in Node.
+if (typeof module !== 'undefined') {
+  module.exports = patchEmitterForWildcard;
+}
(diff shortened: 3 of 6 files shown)


docs/config_file.md
```diff
@@ -66,6 +66,7 @@ Chrome, Chrome Canary, Chromium, Firefox, IE, Opera, PhantomJS, Safari, Safari T
 
 ### Config-level options:
 
+    additionalBrowserSocketEvents[Object]  name of events (keys) to an event handling function
```
step2yeung: suggestion:
`event names (key) to custom event handlers functions to be added to the browser socket`

public/testem/testem_connection.js
```diff
@@ -176,6 +189,11 @@ function initSocket(id) {
   socket.on('stop-run', function() {
     sendMessageToParent('stop-run');
   });
+  socket.on('*', function(event) {
+    var eventName = event.data[0];
```
step2yeung: can we check that event.data is defined?
```
if (event.data) {
    var eventName = event.data[0];
    var eventData = event.data[1];
    sendMessageToParent(eventName, eventData);
}```

File: public/testem/testem_client.js [hunk 3]
```diff
@@ -186,6 +192,20 @@ var Testem = {
...
+    if (!this.evtHandlers) {
+      this.evtHandlers = {};
+    }
+    if (!this.evtHandlers[evt]) {
+      this.evtHandlers[evt] = [];
+    }
+    var callbackIndex = this.evtHandlers[evt].length;
+    var wrapper = function(){
+      callback(...arguments);
+      this.onceCallBackEventIdx.push(callbackIndex);
```
step2yeung: typo: should be `onceCallbackEventIdx`

docs/config_file.md
```diff
@@ -66,6 +66,7 @@ Chrome, Chrome Canary, Chromium, Firefox, IE, Opera, PhantomJS, Safari, Safari T
 
 ### Config-level options:
 
+    additionalBrowserSocketEvents[Object]  event names (key) to custom event handlers functions to be added to the browser socket
```
scalvert: Can we follow the existing naming conventions here? This would become `additional_browser_socket_events`. Additionally, the description is quite confusing - the type is `Object` but you say 'event names (key) to custom event handlers functions...'.

It maybe should be:

> an object containing keys corresponding to event names that point to handler functions, which are to be added to the browser socket

public/testem/testem_client.js
```diff
@@ -127,6 +127,7 @@ var Testem = {
   emitMessageQueue: [],
   afterTestsQueue: [],
   console: {},
+  onceCallbackEventIdx: [],
```
scalvert: This name seems to imply _singular_, but is actually an array of indexes, no? Perhaps we should change to `onceCallbackEventIndexes`?

public/testem/testem_client.js
```diff
@@ -186,6 +192,20 @@ var Testem = {
...
   },
+  once: function(evt, callback) {
+    if (!this.evtHandlers) {
+      this.evtHandlers = {};
+    }
+    if (!this.evtHandlers[evt]) {
+      this.evtHandlers[evt] = [];
+    }
+    var callbackIndex = this.evtHandlers[evt].length;
+    var wrapper = function(){
```
scalvert: Nit: space after `function()`

public/testem/testem_client.js
```diff
@@ -286,6 +309,18 @@ var Testem = {
     // stringify for clients that only can handle string postMessages (IE <= 10)
     return JSON.stringify(message);
   },
+  removeDuplicateCallbacks: function(evt, callback) {
+    var handlers = this.evtHandlers[evt];
+    var removeIdx = [];
```
scalvert: We seem to have a mixture of `let` and `var` in testem. Which is preferred?

public/testem/testem_client.js
```diff
@@ -286,6 +309,18 @@ var Testem = {
...
+    for (var i = 0; i < handlers.length; i++) {
+      if(handlers[i] === callback) {
```
scalvert: Nit: space after `if`

tests/runners/browser_test_runner_tests.js
```diff
@@ -417,6 +417,27 @@ describe('browser test runner', function() {
     });
   });
 
+  describe('additionalBrowserSocketEvents', function() {
+    it('bind events', function() {
```
scalvert: `it('binds additional browser socket events')` maybe?

public/testem/testem_client.js
scalvert: I'd go with `let`, as it's the new syntax.

tests/runners/browser_test_runner_tests.js
```diff
@@ -417,6 +417,27 @@ describe('browser test runner', function() {
...
+      let eventName = 'testEvent';
+      let eventFn = function() {
+        return 'testing additionalBrowserSocketEvents.';
+      }
+      let additionalBrowserSocketEvents = {};
+      additionalBrowserSocketEvents[eventName] = eventFn;
+
+      let socket = new FakeSocket();
+      let reporter = new FakeReporter();
+      let config = new Config('ci', { additionalBrowserSocketEvents, reporter: reporter });
```
step2yeung: should this be 
`let config = new Config('ci', { additional_browser_socket_events: additionalBrowserSocketEvents, reporter: reporter });`

public/testem/testem_client.js
stefanpenner: for browser destined JavaScript, we should not introduce new JS features. We can in the future, but that would require a major version bump for little value.

It is nice, that testem does still support older platforms, dropping support for them should only occur if there is something high value we gain.

docs/config_file.md
```diff
@@ -66,6 +66,7 @@ Chrome, Chrome Canary, Chromium, Firefox, IE, Opera, PhantomJS, Safari, Safari T
 
 ### Config-level options:
 
+    additional_browser_socket_events    [Object]  an object containing keys corresponding to event names that point to handler functions, which are to be added to the browser socket
```
stefanpenner: some bikeshedding, thoughts on:

* `custom_browser_socket_events`
* `browser_socket_events`

public/testem/testem_client.js
```diff
@@ -266,6 +267,9 @@ var Testem = {
         case 'stop-run':
           self.emit('after-tests-complete');
           break;
+        default:
+          self.emit(type, message.data);
```
stefanpenner: emitting all events from window.message as now part of testem's public API seems error prone. Remember, lots of applications actually use these for communication.

Thoughts on only propagating events which we expect rather then all?

One approach, would be to namespace these new events, for example, one could prefix them with `socket-io:${type}` or something, then once we hit this default switch, we only emits those which `startWith('socket-io:')`

public/testem/testem_connection.js
```diff
@@ -176,6 +189,13 @@ function initSocket(id) {
   socket.on('stop-run', function() {
     sendMessageToParent('stop-run');
   });
+  socket.on('*', function(event) {
+    if (event.data) {
+      var eventName = event.data[0];
+      var eventData = event.data[1];
+      sendMessageToParent(eventName, eventData);
```
stefanpenner: I think we should consider namespacing this event, so as to not conflict with other events that application may produce.

Ultimately, I think testem should take this approach for all messages it posts to window

public/testem/testem_client.js
```diff
@@ -266,6 +267,11 @@ var Testem = {
         case 'stop-run':
           self.emit('after-tests-complete');
           break;
+        default:
+          if (type.startsWith('testem:')) {
```
stefanpenner: `startsWith` is ES6 only, so to keep our browser code compat with older versions lets create a `startsWith` helper instead.

I believe the following should do the trick:

```js
function starsWith(str, search, pos) {
  return str.substr(!pos || pos < 0 ? 0 : +pos, search.length) === search;
}
```

choheekim: I made changes to  use `indexOf` for instance`'evnetName.indexOf('testem:') == 0` .

public/testem/testem_client.js
```diff
@@ -266,6 +267,11 @@ var Testem = {
         case 'stop-run':
           self.emit('after-tests-complete');
           break;
+        default:
+          if (type.indexOf('testem:') == 0) {
```
step2yeung: nitpick: `===` instead of `==`

public/testem/testem_connection.js
```diff
@@ -176,6 +189,13 @@ function initSocket(id) {
   socket.on('stop-run', function() {
     sendMessageToParent('stop-run');
   });
+  socket.on('*', function(event) {
+    if (event.data && event.data[0].indexOf('testem:') == 0 ) {
```
step2yeung: nitpick `===` instead of `==`

tests/testem_connection_tests.js
```diff
@@ -0,0 +1,65 @@
...
+
+const patchEmitterForWildcard = require('../public/testem/testem_connection');
+const expect = require('chai').expect;
+const ServerIO = require('socket.io');
+const ClientIO = require('socket.io-client');
+
+describe.only('Testem Connection', function() {
+  var serverIo, client;
+
+  function replaceGlobals(newGlobals, originalGlobals) {
```
step2yeung: I feel we should restore the globals in the after() hook, prevent any gotchas.

tests/testem_connection_tests.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
+const patchEmitterForWildcard = require('../public/testem/testem_connection');
+const expect = require('chai').expect;
+const ServerIO = require('socket.io');
+const ClientIO = require('socket.io-client');
+
+describe.only('Testem Connection', function() {
```
stefanpenner: lets remove the `only`

tests/testem_connection_tests.js
```diff
@@ -0,0 +1,65 @@
...
+    client = ClientIO('http://localhost:8000');
+    patchEmitterForWildcard(client);
+
+    var eventNameArr = [];
+
+    function check(eventName) {
+      eventNameArr.push(eventName);
+      if(eventNameArr.length === 2) {
+        expect(eventNameArr).to.deep.equal(['*', 'foo']);
+        done();
```
stefanpenner: If the above expect fails, does the test suite error report correctly? At one point, mocha would simply let the test timeout here. If you could double check that it does not (maybe share the test output) that would be great.

public/testem/testem_client.js
scalvert: Yes, agree. After speaking offline it appears this was destined for browsers, sometimes older browsers, that don't support newer syntaxes. Let's stick with the old stuff.

tests/testem_connection_tests.js
scalvert: Agree.

tests/testem_connection_tests.js
choheekim: Check if the test failed correctly by changing a value in the array. 
here is the output of test failing purposely.
```
1) Testem Connection patches emitter for wildcard:

      Uncaught AssertionError: expected [ '*', 'foo' ] to deeply equal [ '*', 'baz' ]
      + expected - actual

       [
         "*"
      -  "foo"
      +  "baz"
       ]
```

stefanpenner:
released as v2.14.0 🎉

PR #1314, opened 2018-12-20
Add `tap_errors_only` config option for TAP reporter

jjnypr:
Have you ever been frustrated by having to scroll endlessly through your TAP logs to find the one test failure, and wished Testem TAP reports would only log errors? This PR is for you!

**What changed:** added a new `tap_failed_tests_only` config option, defaulting to false, which, when set to true, suppresses `ok` test results from being sent to `this.out`.

(Not sure why the AppVeyor tests are failing but I assume it's related to them also failing for other recent releases/branches, happy to make make changes if anyone has input.)


step2yeung:
@jjnypr how does it differ from the `tap_quiet_logs` options? The naming of both seems to equate the same functionality
```tap_quiet_logs               [Boolean] whether to suppress non-failing logs in TAP reporting```

**Comment by jjnypr** [comment]
@step2yeung Good question! I initially assumed `tap_quiet_logs` would do this, but after I dove into the source, I saw that what it does is suppress _log_ messages (exceptions, etc), not suppress the logging of passing tests to the console. The difference is clearest in the tests for the two options:

In [reporter_tests.js](https://github.com/testem/testem/blob/e0780e4d77c0db0a56d5784f7db486cd939157cf/tests/ci/reporter_tests.js#L148-L183), you can see the tests confirm that the logs aren't present in the output when `tap_quiet_logs` is true:

```javascript
    context('with quiet logs', function() {
      beforeEach(function() {
        config = new Config('ci', { tap_quiet_logs: true });
      });


      context('without errors', function() {
        it('writes out TAP', function() {
          var reporter = new TapReporter(false, stream, config);
          reporter.report('phantomjs', {
            name: 'it does stuff',
            passed: true,
            logs: ['some log'],
            runDuration: 3,
          });
          reporter.report('phantomjs', {
            name: 'it is skipped',
            skipped: true,
            logs: [],
            runDuration: 0,
          });
          reporter.finish();
          assert.deepEqual(stream.read().toString().split('\n'), [
            'ok 1 phantomjs - [3 ms] - it does stuff',
            'skip 2 phantomjs - [0 ms] - it is skipped',
            '',
            '1..2',
            '# tests 2',
            '# pass  1',
            '# skip  1',
            '# fail  0',
            '',
            '# ok',
            ''
          ]);
        });
      });
```

The tests for this PR confirm instead that _only_ failing tests are written to the console, and don't affect the log output captured during the test. You can see how `tap_quiet_logs` is used in [displayutils.js](https://github.com/testem/testem/blob/1960a16dd5de3611910f96a00b2b86da8fd39fdb/lib/utils/displayutils.js#L46-L52):

```javascript
function resultString(id, prefix, result, quietLogs) {
  let string = resultDisplay(id, prefix, result) + '\n';
  if (result.error || (!quietLogs && result.logs && result.logs.length)) {
    string += yamlDisplay(result.error, result.logs) + '\n';
  }
  return string;
}
```

In this case, if there _is_ and error _or_  `tap_quiet_logs` is `false` (the default) console output (deprecation warnings, etc) will be logged. That's useful but I think there's a good case to be made that only writing _errors_ to the output is distinct and also useful, especially if you're reviewing a failed build in a CI environment.

jjnypr:
Would it help if I renamed the option to something like `tap_test_failures_only`?

step2yeung:
Apologies, forgot to reply to this.

@jjnypr Nope, I think the original naming is fine! But could you update the description of `tap_quiet_logs` with what you found to make it more clear of what it does and doesn't do?
Otherwise, I think this looks good.

jjnypr:
@step2yeung Modified descriptions to make the difference clearer, without getting too verbose. I renamed my new config option anyway, `tap_errors_only` feels much less clear than `tap_failed_tests_only`, which is pretty explicit.

jjnypr:
@step2yeung is there a good way to re-run the CI builds? When I skimmed the results it looks like these were transient failures.

step2yeung:
@jjnypr Thanks for the ping! Merged!

PR #1335, opened 2019-04-07
Fix appveyor test: failing due to null this.process

step2yeung:
Appveyor fails on `dev mode app in interactive mode allows to restart a run` test.

Progress so far:
The test invokes 3 browsers, and will restart the run and assert the browsers will be killed.
All the browsers fires the `closed` event that sets `this.process = null`.
But for the node instance, the onKillTimeout is set and expires, which tries to access this.process after it has been set to null.

### DIFF
--- a/lib/utils/process.js
+++ b/lib/utils/process.js
@@ -110,7 +110,15 @@ module.exports = class Process extends EventEmitter {
         log.error('Error killing process ' + self.name + '.', err);
       });
       self._killTimer = setTimeout(self.onKillTimeout.bind(self), self.killTimeout);
-      kill(self.process, sig);
+
+      // Check for self.process to make sure the process is still available.
+      // Otherwise don't do anything.
+      // See https://github.com/testem/testem/pull/1335 for more information.
+      if (self.process) {
+        kill(self.process, sig);
+      } else {
+        log.info('Process ' + this.name + ' already killed.');
+      }
     });
   }
 };


jasonmit:
@step2yeung is this related to any specific issues?

I only ask because I'm tackling testem + appveyor integration within our project pollyjs and noticing appveyor isn't waiting on my test run or quietly exits (unsure which right now).  Curious if it's related.  Thanks in advance.

step2yeung:
@jasonmit this PR is specifically to address tests that are flakey, giving us a noisy signal when we try to push to master.
I haven't looked at this deep enough to say if its test code or if its appveyor related. I can give you a ping once I find out more

step2yeung:
Sharing my understanding of the problem:

Some background:
When a process is terminaled, process.kill is called and sets up `self._killTimer` which will call `kill()` on line 137 of process.js. I believe this is done incase the process isn't killed properly.
The kill() function is passed the process instance and the signal, and on the windows codepath, it will check for process.pid, and try to kill the process

The problem:
In the test that is failing, `stopRunners()` and `killRunners` is both called from `app.js`. Both will go thru `TapProcessTestRunner.exit` to call `Process.kill`
The stack trace here illustrates this: https://ci.appveyor.com/project/johanneswuerbach/testem/builds/26833367

In this instance, Node's process has killed called twice: once by `stopRunners()`, and once by `killRunners()`. This means there are two timeout set to kill the Node process.
When the Node process is killed, it clears one of the timeout. Leaving the other timeout hanging.
Once the other timeout expires, it will call kill on the Node process which has already been killed. Hence we get the failure:
```
Uncaught TypeError: Cannot read property 'pid' of null
      at kill (lib\utils\process.js:147:33)
      at Process.onKillTimeout (lib\utils\process.js:39:5)
```

The current solution to this is to check for whether `this.process` exists before calling kill() on it.
This does not seem to be a problem on linux.

To understand it further, we need to know why both stopRunners() and killRunners() gets called. If this is a legitimate flow that we should account for with the solution above or is there something deeper.

@stefanpenner Thoughts?

johanneswuerbach:
Thanks for the investigation 👏 

If I recall correctly `stopRunners` was a graceful stop and `killRunners` a more forceful termination (think SIGTERM and SIGKILL in Linux) and both are do via `kill()`

As all of this code is async doing this process check only here https://github.com/testem/testem/blob/master/lib/utils/process.js#L88 is probably racy and we should add the check you proposed.

step2yeung:
Travis-ci still fails due to #1352, this is known.

@johanneswuerbach When you have time, mind taking a look at the fix?

PR #1338, opened 2019-05-05
Cleanup: Moving utils files into lib/utils

step2yeung:
Moving utils files into lib/utils


step2yeung:
@johanneswuerbach would you have time to have a look? I thought the above would clean up `/lib` a bit

johanneswuerbach:
@step2yeung looks great, thank you 👍

PR #1345, opened 2019-06-14
docs/custom_reporters: Improve code snippets

Turbo87:
These changes ensure that the right syntax highlighting is used, and transform the code snippets to use more modern JS syntax


johanneswuerbach:
👍

PR #1364, opened 2019-08-13
Add browser-relogin event

step2yeung:
_**Problem:**_
The recently added reconnect event handler in #1340 introduced a bug that would cause browser timeout errors due.

This is due the code to find existing browserTestRunner in `onBrowserLogin()` that checks for:
`(!runner.socket || !runner.socket.connected);`

The original assumption was the browserTestRunner would have `runner.socket.connected === false` when a reconnection happens, but that is not the case. Instead, `runner.socket.connected === true` and a new browserTestRunner will be created. This leaves the existing browserTestRunner to hang around, causing a browser timeout error.

**Solution:**
Add a new `browser-relogin` event that is fired on socket reconnection. This event check's existing browserTestRunner whether it had a `runner.socket` or if `runner.socket === null` which is set by `onDisconnect()`.
This allows reuse of existing browserTestRunner.

- [x] Add tests (no tests currently for onBrowserLogin or onBrowserRelogin)

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -226,6 +226,7 @@ module.exports = class App extends EventEmitter {
     this.server = new Server(this.config);
     this.server.on('file-requested', this.onFileRequested.bind(this));
     this.server.on('browser-login', this.onBrowserLogin.bind(this));
+    this.server.on('browser-relogin', this.onBrowserRelogin.bind(this));
     this.server.on('server-error', this.onServerError.bind(this));
 
     return this.server.start().asCallback(callback);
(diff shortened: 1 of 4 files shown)


step2yeung:
Added tests. 

AppVeyor test failing for the same failures that is being investigated in #1335 
cc @stefanpenner

stefanpenner:
@step2yeung :shipit: 

If you have some time, could you also investigate/stabilize the tests?

step2yeung:
Merging this, although more of the appveyor inconsistent failures needs some investigating

PR #1372, opened 2019-09-16
[WIP] Adding edge beta to known-browsers

step2yeung:
For those who wants to use edge beta from https://www.microsoftedgeinsider.com/en-us/

TODO:
- verify which chrome switches applies to edge

Note1: Tested with no switches, edge does not start up.
Note2: Tested with headlessChromeArgs, edge does not start up.


step2yeung:
Closing since #1410 landed :)

PR #1385, opened 2019-11-13
adding willDisplay() so subclasses can check that before trying to display()

davecombs:
in case they have side-effects


step2yeung:
@davecombs could we add a test for this?

davecombs:
Oops - yes, I will definitely do that now. Thanks, Stephen!

step2yeung:
Looks good! Thanks @davecombs

Issue #1352, opened 2019-07-11
Trunk is red due to issue with saucelabs


stefanpenner:
https://travis-ci.org/testem/testem/jobs/557407168

```
not ok 3 SL_Chrome_Current_No_Details - [undefined ms] - bailout
    ---
        message: >
            Error: [init({"browserName":"chrome","deviceName":"","deviceOrientation":"","tags":["Saucie","test"],"name":"Saucie tests","public":"public","build":"3418","tunnelIdentifier":"3418.5","version":"latest","platform":"Windows 10"})] The environment you requested was unavailable.
    ...
not ok 4 SL_Chrome_Current - [undefined ms] - error
    ---
        message: >
            Error: Browser exited unexpectedly
            Stderr: 
             { Error: [init({"browserName":"chrome","deviceName":"","deviceOrientation":"","tags":["Saucie","test"],"name":"Saucie tests","public":"public","build":"3418","tunnelIdentifier":"3418.5","version":"latest","platform":"Windows 10"})] The environment you requested was unavailable.
            Sauce could not start your job. For more information on what happened, please visit https://app.saucelabs.com/tests/d4e749d8da274d5db32712b9f4285233
                at /home/travis/build/testem/testem/node_modules/wd/lib/webdriver.js:140:15
                at Request._callback (/home/travis/build/testem/testem/node_modules/wd/lib/http-utils.js:89:7)
                at Request.self.callback (/home/travis/build/testem/testem/node_modules/wd/node_modules/request/request.js:186:22)
                at emitTwo (events.js:126:13)
                at Request.emit (events.js:214:7)
                at Request.<anonymous> (/home/travis/build/testem/testem/node_modules/wd/node_modules/request/request.js:1163:10)
                at emitOne (events.js:116:13)
                at Request.emit (events.js:211:7)
                at IncomingMessage.<anonymous> (/home/travis/build/testem/testem/node_modules/wd/node_modules/request/request.js:1085:12)
                at Object.onceWrapper (events.js:313:30)
                at emitNone (events.js:111:20)
                at IncomingMessage.emit (events.js:208:7)
                at endReadableNT (_stream_readable.js:1064:12)
                at _combinedTickCallback (internal/process/next_tick.js:139:11)
                at process._tickCallback (internal/process/next_tick.js:181:9)
              cause: 
[...]

PR #1386, opened 2019-11-28
[BugFix] On browserRelogin, clear timeouts if socket is not null & log.warn instead of error for reconnection from unknown browser

step2yeung:
**Problem:**

1. On browserRelogin event, there are times where the socket is not null. When we call tryAttach() on an existing socket, eventListeners attached from `tryAttach()` will be duplicated. 

2. When there is an unknown browser reconnecting, app.js will throw an error and the run will fail. This could happen if a previous run's browser for some reason did not terminate. 

**Solutions:**
1. On browserRelogin, clear timeouts if socket is not null. This will prevent socket events duplication.

2. `log.warn` instead of throwing an error for reconnection from unknown browser. This way, the current run won't be punished for dangling browsers from previous runs.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -281,7 +281,11 @@ module.exports = class App extends EventEmitter {
       return;
     }
 
-    browser.tryAttach(browserName, id, socket);
+    if (browser.socket !== null) {
+      browser.clearTimeouts();
+    } else {
+      browser.tryAttach(browserName, id, socket);
+    }
   }
 
   addRunner(runner) {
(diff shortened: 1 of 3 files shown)


step2yeung:
Travis failure due to sauce lab issues: #1352

step2yeung:
@stefanpenner Would love to get your eyes on this PR too!

stefanpenner:
@step2yeung LGTM – :shipit:

PR #1398, opened 2020-02-14
Add option for strict TAP compliance

wagenet:
Currently includes #1399 so tests will run. Should be rebased once merged.

---

This resolves the debate at https://github.com/testem/testem/pull/1101#discussion_r108010346 by side-stepping the issue.


wagenet:
Still needs tests.

PR #1399, opened 2020-02-14
Remove `it.only` that should have been for local testing only

wagenet:
Incorrectly introduced in 6c714be5faf4ed50e7366be08b4da250ff0df4e2


wagenet:
Looks like this uncovers an issue that was missed.

step2yeung:
@wagenet discarding this PR since the changes was merged as part of your previous change.

PR #1400, opened 2020-02-14
QUnit.todo Support

wagenet:
Currently includes #1398 and #1399. Should be rebased once those are merged.

---

Supersedes #1101

Includes support for TAP, Dot, and dev reporters. May not work properly with other options.

### DIFF
--- a/examples/qunit_lazy/tests.js
+++ b/examples/qunit_lazy/tests.js
@@ -5,3 +5,13 @@ QUnit.test('says hello world', function(assert){
 QUnit.test('says hello to person', function(assert){
     assert.equal(hello('Bob'), 'hello Bob', 'should equal hello Bob');
 });
+
+QUnit.todo('failing todo', function(assert){
+    assert.ok(false, 'should be true');
+});
+
+// This will cause integration tests to fail because QUnit will report passing todos
+// as failures (since you should just mark them as a normal test if they pass).
+// QUnit.todo('passing todo', function(assert){
+//     assert.ok(true, 'should be true');
+// });
--- a/lib/reporters/dot_reporter.js
+++ b/lib/reporters/dot_reporter.js
@@ -12,6 +12,7 @@ module.exports = class DotReporter {
     this.total = 0;
     this.pass = 0;
     this.skipped = 0;
+    this.todo = 0;
     this.results = [];
     this.startTime = new Date();
     this.endTime = null;
@@ -30,8 +31,10 @@ module.exports = class DotReporter {
     this.total++;
     if (data.skipped) {
       this.skipped++;
-    } else if (data.passed) {
+    } else if (data.passed && !data.todo) {
       this.pass++;
+    } else if (!data.passed && data.todo) {
+      this.todo++;
     }
   }
 
@@ -43,8 +46,10 @@ module.exports = class DotReporter {
       this.currentLineChars = 0;
       this.out.write('\n  ');
     }
-    if (result.passed) {
+    if (result.passed && !result.todo) {
       this.out.write('.');
+    } else if (!result.passed && result.todo) {
+      this.out.write('T');
     } else if (result.skipped) {
       this.out.write('*');
     } else {
--- a/tests/ci/reporter_tests.js
+++ b/tests/ci/reporter_tests.js
@@ -332,37 +332,37 @@ describe('test reporters', function() {
     });
 
     context('willDisplay', function() {
-      it.only('silent only, no result', function() {
+      it('silent only, no result', function() {
         config = new Config('ci', {});
         var reporter = new TapReporter(true, stream, config);
         assert.equal(reporter.willDisplay(), false);
       });
-      it.only('silent only, has result with no error', function() {
+      it('silent only, has result with no error', function() {
         config = new Config('ci', {});
         var reporter = new TapReporter(true, stream, config);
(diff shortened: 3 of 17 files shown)


wagenet:
I've tested this against an Ember app using QUnit.todo and it works well enough for me to want to use it. Of course, I understand if there's a higher bar for merging :)

lib/reporters/dot_reporter.js
```diff
@@ -43,8 +46,10 @@ module.exports = class DotReporter {
       this.currentLineChars = 0;
       this.out.write('\n  ');
     }
-    if (result.passed) {
+    if (result.passed && !result.todo) {
       this.out.write('.');
+    } else if (!result.passed && result.todo) {
```
step2yeung: can probably just have `else if(result.todo) {` here since `result.passed` would be false (goes towards the rest fo the changes in the PR)

lib/reporters/dot_reporter.js
```diff
@@ -43,8 +46,10 @@ module.exports = class DotReporter {
       this.currentLineChars = 0;
       this.out.write('\n  ');
     }
-    if (result.passed) {
+    if (result.passed && !result.todo) {
```
step2yeung: Help jog my memory.. is a `todo` reported as pass as well? If not, probably wouldn't need `!result.todo` ? (goes towards the rest fo the changes in the PR)

examples/qunit_lazy/tests.js
```diff
@@ -5,3 +5,12 @@ QUnit.test('says hello world', function(assert){
...
 });
+
+QUnit.todo('failing todo', function(assert){
+    assert.ok(false, 'should be true');
+});
+
+// This will cause integration tests to fail
+// QUnit.todo('passing todo', function(assert){
+//     assert.ok(true, 'should be true');
+// });
```
step2yeung: Do we still need this?

wagenet: I can take it out. I had it in there originally because I wanted to test that that case behaved correctly, but the integration test runner expects the tests to all pass. I didn't see a mechanism for intentional failure.

lib/reporters/dot_reporter.js
wagenet: For QUnit a todo can either pass or fail. A pass means all assertions passed, even though it was marked as todo. QUnit's UI treats this as a failure, even though internally it's marked as passing, because you should stop marking it as a todo if it passes.

lib/reporters/dot_reporter.js
wagenet: Per my previous comment, you can have a passing todo in QUnit. This is considered a bad thing, since you should mark it as a regular test, not a todo anymore.

wagenet:
@step2yeung any thoughts?

lib/reporters/dot_reporter.js
step2yeung: Thanks for the summary!

examples/qunit_lazy/tests.js
step2yeung: Cleaning this up would be great :)

lib/reporters/dot_reporter.js
step2yeung: resolving this as part of the other comment

examples/qunit_lazy/tests.js
wagenet: @step2yeung I've updated the comment.

step2yeung:
@wagenet thanks for the ping & your patiences! Resolved 2 of the comments, also just pulling in the pr myself to see how it works :)

tests/ci/reporter_tests.js
```diff
@@ -221,6 +224,48 @@ describe('test reporters', function() {
...
             '# skip  1',
+            '# todo  0',
+            '# fail  1',
+            ''
+          ]);
+        });
+      });
+
+      context('with todos', function() {
+        it('writes out TAP with failure info', function() {
```
step2yeung: This test is failing for me when checking out the branch locally:
```
 1) test reporters tap reporter with error-only output with todos writes out TAP with failure info:

      AssertionError: expected [ Array(16) ] to deeply equal [ Array(22) ]
      + expected - actual

         "        browser log: |"
         "            I am a log"
         "            Useful information"
         "    ..."
      +  "not ok 2 phantomjs - [5 ms] - it is a passing todo"
      +  "    ---"
      +  "        browser log: |"
      +  "            I am a log"
      +  "            Useful information"
      +  "    ..."
         ""
         "1..2"
         "# tests 2"
         "# pass  0"

      at Context.<anonymous> (tests/ci/reporter_tests.js:400:18)
      at processImmediate (internal/timers.js:439:21)
```

Mind taking a look?

wagenet: I believe I've fixed the issue.

wagenet:
@step2yeung I believe the remaining failures are not my fault.

tests/ci/reporter_tests.js
step2yeung: Thanks for the fix!

step2yeung:
@wagenet merged! Thanks for wrapping up the PR!

fyi @rwjblue

rwjblue:
Awesome, thank you both!  

@step2yeung - Mind shipping a release?

step2yeung:
@rwjblue @wagenet released as v3.0.3 !

PR #1402, opened 2020-04-20
Add Brave to known browsers

### LINKED ISSUE #1339: Add support for Brave browser
Would be nice if https://brave.com browser was also available in known browser config: https://github.com/testem/testem/blob/47773ffe85627d7486496e5999082b01b4add222/lib/utils/known-browsers.js

ming-codes:
Resolves #1339

I only tested on MacOS. The Windows path is taken from elsewhere, but I figured it should be correct.


johanneswuerbach:
Released as `v3.1.0` :-)

PR #1421, opened 2021-03-15
fix: CVE-2021-21366

artemgurzhii:
https://www.npmjs.com/advisories/1650

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "styled_string": "0.0.1",
     "tap-parser": "^7.0.0",
     "tmp": "0.0.33",
-    "xmldom": "^0.1.19"
+    "xmldom": "^0.6.0"
   },
   "files": [
     "lib",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5474,9 +5474,9 @@
       }
     },
     "xmldom": {
-      "version": "0.1.27",
-      "resolved": "https://registry.npmjs.org/xmldom/-/xmldom-0.1.27.tgz",
-      "integrity": "sha1-1QH5ezvbQDr4757MIFcxh6rawOk="
+      "version": "0.6.0",
+      "resolved": "https://registry.npmjs.org/xmldom/-/xmldom-0.6.0.tgz",
+      "integrity": "sha512-iAcin401y58LckRZ0TkI4k0VSM1Qg0KGSc3i8rU+xrxe19A/BN1zHyVSJY7uoutVlaTSzYyk/v5AmkewAP7jtg=="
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
     "styled_string": "0.0.1",
     "tap-parser": "^7.0.0",
     "tmp": "0.0.33",
-    "xmldom": "^0.1.19"
+    "xmldom": "0.5.0"
```
patocallaghan, line 52: Should this be `^0.5.0` instead considering the existing dep has `^`?

johanneswuerbach:
The test failures look related, could you take a look?

artemgurzhii:
@johanneswuerbach I have fixed main failing tests, but few browser tests keep failing, also I have noticed that it happens on the master branch and others, do I need to fix it or it's ok?

tests/ci/reporter_tests.js
```diff
@@ -768,7 +768,7 @@ describe('test reporters', function() {
       reporter.finish();
       var output = stream.read().toString();
       assert.match(output, /<testsuite name="Testem Tests" tests="1" skipped="0" failures="0" timestamp="(.+)" time="(\d+(\.\d+)?)">/);
-      assert.match(output, /<testcase classname="phantomjs" name="it does &lt;cool> &quot;cool&quot; 'cool' stuff"/);
+      assert.match(output, /<testcase classname="phantomjs" name="it does <cool> &quot;cool&quot; 'cool' stuff" time="0"\/>/);
```
johanneswuerbach, line 771: Shouldn't this actually be escaped? If I understand the test correctly having a `<` here would mean invalid XML.

artemgurzhii: no idea actually, I have just fixed tests 🙂

johanneswuerbach: It looks like this is actually a regression in 0.5.0 and we need to wait for 0.6.0 https://github.com/xmldom/xmldom/pull/199

johanneswuerbach: Those changes should be reverted now :-)

artemgurzhii: done, do I need to merge commits into one?

johanneswuerbach: That would be appreciated, otherwise 👍

artemgurzhii: done!

johanneswuerbach:
Thank you 🙇 , published as `3.4.1`.

PR #1471, opened 2021-08-11
Fix bail_on_uncaught_error

ef4:
`bail_on_uncaught_error` currently does not cause testem to shut down after a top-level error.  So if a top-level error causes your test suite to be dead, testem will hang forever.

It looks like this feature was broken by  https://github.com/testem/testem/commit/a690bfa41708b9ef2903c34cee214ba70862e88d

Prior to that change, it was sufficient for the error handler to call `onAllTestResults()`, because `onAllTestResults` invoked `finish()`. 

This PR restores the behavior of calling `finish()` when trying to bail.

And the error handler doesn't need to call `onEnd` because that's already called by `onAllTestResults`

### DIFF
--- a/lib/runners/browser_test_runner.js
+++ b/lib/runners/browser_test_runner.js
@@ -248,7 +248,7 @@ module.exports = class BrowserTestRunner {
         error: {}
       });
       this.onAllTestResults();
-      this.onEnd.call(this);
+      this.finish();
     }
   }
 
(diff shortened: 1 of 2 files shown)


johanneswuerbach:
Could you add a test to ensure we don't regress again?

ef4:
I added test coverage.

rwwagner90:
Really looking forward to this, so my GitHub actions builds won't all take 360 minutes to fail 😃

johanneswuerbach:
Thank you, published as `testem@3.5.0`

PR #1482, opened 2021-11-08
Replace fireworm with sane

### LINKED ISSUE #1481: fireworm is pollutting logs
`yarn why` points at `testem @ 0.7.1` as being the hosting package for fireworm in my case.

getting a bunch of 
```
DeprecationWarning: 
  Invalid 'main' field in `<absolute path>/node_modules/fireworm/package.json' of 'fireworm.js'. 
  Please either fix that or report it to the module author.

```

mydea:
This PR replaces the pretty old [fireworm](https://github.com/airportyh/fireworm) dependency with [sane](https://github.com/amasad/sane).

The actual watching behavior was not really tested that well, so I tried to add some more tests to cover if it actually picks up changing files, and if the core configuration works as expected. I _think_ the watching used to ignore the `cwd` setting (if I understood everything correctly), which it now shouldn't.

To test the file watching, I had to add a timeout to the test.. doesn't feel super nice, but I wasn't sure how else to properly test this. Happy to change this if there is a better way to do this! 

Closes https://github.com/testem/testem/issues/1481


RuslanZavacky:
@johanneswuerbach do you think you might have time to look into that if it is good to be merged? :)

johanneswuerbach:
Hey sorry, I've retriggered CI, which would need to be green before merging this (except the browser-tests jobs).

mydea:
Hmm, interesting. I am not really sure what the issue is there - it seems to be some kind of weird (?) timing issue. Not 100% sure if it is actully an introduce "failure" or some other thing. 

When trying to debug it, I for example increased the timeout here:
https://github.com/testem/testem/blob/master/tests/api_tests.js#L131

which also lead to the test failing on master as well (e.g. set it to 1000, and tests will fail). Which, if I understand the test correctly, shouldn't be the case? So I could it be it actually just shows some fundamental timing issue with the test here, maybe? But not 100% sure how all of that plays together, to be honest...!

NullVoxPopuli:
anything that I can do to help move this along?

gnclmorais:
@NullVoxPopuli and others, I forked @mydea’s branch, tweaked the GitHub Actions configuration a bit, and I was able to surface the following errors that we have to look into: https://github.com/gnclmorais/testem/runs/5611283650

johanneswuerbach:
This PR (or a new version of this) would need to be updated and tests passing, afterwards I'm happy to merge this.

johanneswuerbach:
Replaced by https://github.com/testem/testem/pull/1981

PR #1566, opened 2022-06-07
Support Microsoft Edge in Linux

kellyselden:


kellyselden:
Thanks for the quick turnaround!

PR #1580, opened 2022-08-26
Resolve a promise when returned in config

jrjohnson:
This unlocks doing some async work as part of the configuration. For
example using our browserlist query string and the browserstack API to
generate a list of launchers so we're always testing against our
supporter browser matrix.


johanneswuerbach:
lgtm, but there seems to be some linting issue. Could you have a look @jrjohnson ?

jrjohnson:
Whoops, sorry @johanneswuerbach - fixed now.

johanneswuerbach:
Thank you, released as `testem@3.9.0`

PR #1785, opened 2024-04-11
Apply middleware before all served assets.

brettburley:
This enables configured middleware to apply to all served assets, including those served from the public/testem folder.

This is intended to address https://github.com/testem/testem/issues/1784.

### DIFF
--- a/lib/server/index.js
+++ b/lib/server/index.js
@@ -138,6 +138,8 @@ class Server extends EventEmitter {
 
     this.configureProxy(app);
 
+    app.use('/testem', express.static(`${__dirname}/../../public/testem`));
+
     app.get('/', (req, res) => {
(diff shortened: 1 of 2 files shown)


johanneswuerbach:
Hey, thanks for the contribution. Would you mind to add a test here https://github.com/testem/testem/blob/master/tests/server_tests.js#L116 that ensures we don't regress? Maybe use a static middleware that adds an additional response header to all responses and check that is added on static files?

brettburley:
Thanks, good call! I went ahead and added test assertions for the middleware to all of the request types where middleware should be applied. Let me know if you have any feedback.

johanneswuerbach:
Thank you @brettburley , published as `testem@3.13.0`

PR #1814, opened 2024-06-26
🔥 remove package-lock, add .npmrc

ctcpip:


ctcpip:
that is... unexpected.  a regression caused by an unexpected non-semver breaking change where xml `>` was not previously converted to `&gt;`, so probably the fix here is a test change

ctcpip:
yep: 

https://github.com/xmldom/xmldom/blob/master/CHANGELOG.md#082

ctcpip:
alright, tests fixed and all passing now. should be good to merge.

re: dependabot: not sure you would want more noise than necessary.  😄  it still creates PRs for major bumps

PR #1838, opened 2024-09-10
fix: Provide more context when launcher is not available

MichalBryxi:
- Add information which launchers are available
- Helps with #1837
- Example of new error message:
```
Launcher IE not found. Not installed? Available: safari
```

PR #1840, opened 2024-09-11
fix: Make sure that multi-word launcher names are not confused

MichalBryxi:
- Without this fix we can get confusing messages like: `Launcher Chrome not found. Not installed? Available: chrome dev, safari`, which can read like three browsers "chorme", "dev" and "safari".
- By adding surrounding quotes the message reads better and is less ambiguous.
- Also added a full stop to make the message consistent with the others.
- Example of the message now: `Launcher IE not found. Not installed? Available: "chrome", "headless chrome", "safari".`
- Improves #1838


johanneswuerbach:
Released as `testem@3.15.2`, thank you for your contribution 🤩

PR #1886, opened 2025-05-11
support testem.cjs, for use in ESM / (type=module) projects

NullVoxPopuli:
the default config, `testem.js`, is interpreted as an ES Module, when the package.json type=module is set.

To use testem, we still need a cjs config file -- so support the `.cjs` extension makes sense.

Been testing this over here: https://github.com/NullVoxPopuli/nullui/pull/35

via running:
```bash
~/Development/NullVoxPopuli/testem/testem.js
#                           ^ my clone of this repo
#                                  ^ bin file
```

### DIFF
--- a/lib/server/index.js
+++ b/lib/server/index.js
@@ -151,6 +151,9 @@ class Server extends EventEmitter {
     app.get('/testem.js', (req, res) => {
       this.serveTestemClientJs(req, res);
     });
+    app.get('/testem.cjs', (req, res) => {
+      this.serveTestemClientJs(req, res);
+    });
 
     app.all(/^\/(?:-?[0-9]+)(\/.+)$/, serveStaticFile);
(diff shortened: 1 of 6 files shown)


lib/server/index.js
```diff
@@ -151,6 +151,9 @@ class Server extends EventEmitter {
     app.get('/testem.js', (req, res) => {
       this.serveTestemClientJs(req, res);
     });
+    app.get('/testem.cjs', (req, res) => {
```
johanneswuerbach, line 154: This shouldn't be needed as the `cjs` config file should only be read by the server and not be exposed to the client.

NullVoxPopuli: ah nice -- I got confused by the file + path being named the same. I just verified this locally. Will remove.

NullVoxPopuli: updated

PR #1947, opened 2026-03-22
Upgrade execa to v9.6.1

### LINKED ISSUE #1937: testem@3.16.0 depends on execa@1.0.0, which is flagged as vulnerable (command injection vulnerability)
My dependency scanner is flagging a security vulnerability coming from Testem’s dependency on `execa@1.0.0`, which is known to be affected by a command injection issue when used with `shell: true` or `execa.shell()`.

**Dependency chain:**
`ember-cli@4.12.3`
└── `testem@3.16.0`
    └── `execa@1.0.0`  ← vulnerable version

Running `npm ls execa` confirms that Testem is the primary source of the vulnerable version in my project.

**Why this matters:**

`execa` <2.0.0 is flagged as vulnerable by multiple scanners (GitHub Advisory, Snyk, etc.).

Most of the ecosystem has already upgraded to `execa` v2+ or v5+.

Testem’s usage of execa predates the API changes, so projects depending on Ember CLI 4.x cannot override it safely without breaking Testem.

**Request:**

Would it be possible to update Testem’s dependency to a non-vulnerable version of `execa` (e.g., `execa` >= 2.0.1 or ideally a modern version)?

If API changes require small adjustments, I’d be happy to help test a branch or submit a PR.

**Environment:**

`testem: 3.16.0`

`ember-cli: 4.12.3`

`Node: v18.16.0`

`OS: Debian GNU/Linux 11 (bullseye)`

Thanks for your time and for maintaining Testem!

Gaurav0:
Fixes #1937 

See passing tests: https://github.com/Gaurav0/testem/pull/6

`execa` does not call `process.on('error', ...)` any more.
`execa` reports a failure if a process has any exit code other than zero
`execa` cannot reliably determine if a process failed to run normally but has a nonzero exit code or it failed to run in an exceptional way.

if a process failed to spawn or did not start because:
  - it was not found
  - it does not have permissions to run

`execa` fails just as it would if the process spawned but exited with a non-zero exit code. I have resorted to error message parsing to make the tests pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 npm-debug.log
 npm-debug.log.*
 testem.log
+package-lock.json
+examples/*/package-lock.json
--- a/bin/run-integration.js
+++ b/bin/run-integration.js
@@ -61,7 +61,7 @@ function shellExec(cmd, runOpts) {
     return shell.exec(cmd, runOpts, function(exitCode, output) {
       if (exitCode !== 0) {
         return callback(new Error(
-          'Cmd: ' + cmd + ' failed with exit code: ' + exitCode + '\n' + output
+          'Cmd: ' + cmd + ' in directory: ' + path.basename(runOpts.cwd) + ' failed with exit code: ' + exitCode + '\n' + output
         ));
       }
       callback(null, output);
--- a/lib/process-ctl.js
+++ b/lib/process-ctl.js
@@ -1,7 +1,7 @@
 
 
 const log = require('npmlog');
-const execa = require('execa');
+const execa = require('execa').execa;
 const Bluebird = require('bluebird');
 const util = require('util');
 const EventEmitter = require('events').EventEmitter;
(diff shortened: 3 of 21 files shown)


mansona:
@johanneswuerbach this happens because execa is now an ESM only package and require(esm) was only added in Node 20.19

This means that updating execa like this makes it so that anyone consuming testem would need to update their node version (if below 20.19) so it makes this a breaking change.

I strongly recommend that you revert this change, release it as a patch and then release a new major with the updated execa (if you want to do that)

johanneswuerbach:
We don't have the capacity to test against every minor node version and testem doesn't actively support old minor versions.

While this breaking change is unfortunate, we won't rollback in this case. If you need to work with an old nodejs version, please pin your testem version.

mansona:
This isn't just about minor versions though 🤔 require(esm) was **backported** to Node 20.19, which is why we're specifically talking about that version. Before this change testem worked in Node 18 but this change breaks that and should have gone out as a breaking change.

Just to be clear, testem 3.17 worked perfectly fine in Node 18 and some (or all) of the recent changes have broken Node 18 support and thus should have gone out in a Major version and not a minor

johanneswuerbach:
As said, we only support the latest minors across 20/22/24, if a different version worked before, this was mainly supported by accident and not a deliberate choice.

Therefore we don't consider this as breaking change.

PR #1981, opened 2026-04-04
[BREAKING] Replace fireworm with chokidar 5

Gaurav0:
## Summary

**BREAKING CHANGE.** Replaces the **fireworm**-based filesystem watcher with **[chokidar](https://github.com/paulmillr/chokidar) v5** (`^5.0.0`), tightens supported **Node.js** versions, and documents user-facing watch behavior plus troubleshooting. Config keys `src_files`, `watch_files`, `src_files_ignore`, and `disable_watching` are unchanged, but some environments may see different watch behavior than with fireworm (see **Breaking changes**).

### Node.js (`engines` in `package.json`)

Supported Node was previously **`>= 20.*`**, which (under normal semver checks) allows **any** Node **≥ 20**, including **21**, **23**, and other non-LTS majors. It is now restricted to **LTS-aligned ranges with minimum patch versions**:

`^20.19.0 || ^22.12.0 || >= 24.0.0`

That **drops** odd major lines (e.g. **21**, **23**) and versions below the stated minimums on the 20 / 22 lines. Consumers on unsupported Node versions must upgrade Node (or stay on an older Testem release).

### Fireworm → chokidar (file watching)

Swapping the file watch backend is a **behavioral** breaking change even when config is identical:

- **OS / filesystem differences:** Native watchers and polling differ by platform; chokidar’s defaults and options are not the same as fireworm’s. Events may differ in timing, batching, or whether certain operations (e.g. some **rename** / **replace** flows, **symlinks**, or **networked** paths) surface the same way.
- **Flaky or virtualized trees:** Paths that only worked with fireworm may need **`CHOKIDAR_USE_POLLING=1`** or **`CHOKIDAR_INTERVAL`** (documented in the README and `docs/config_file.md`).
- **Intent:** Testem still applies the same **glob policy** (include/ignore from config) on top of directory watching so *configured* semantics stay aligned; the remaining risk is in **low-level** event delivery, not in renaming config options.

If you rely on very specific fireworm-era quirks, validate dev-mode watching on your OS and CI after upgrading.

## Motivation

- **Maintenance:** The prior watcher dependency, **fireworm**, is **no longer maintained**, which increases long-term risk (security, compatibility, and the cost of fixes). Moving to an **actively maintained** replacement is a prerequisite for sustainable releases.
- **Why chokidar instead of [sane](https://github.com/amasad/sane)?** An obvious alternative was to follow the pattern other projects use and adopt **sane** (a small wrapper around watchers). That path was considered—including existing community work to swap fireworm for sane—but **sane’s own maintenance** has been **sporadic** for years, so trading one low-activity dependency for another did not improve the long-term story. **Chokidar v5** is **actively maintained**, widely deployed, and is what **modern dev tooling** relies on for file watching: for example **[Vite](https://vitejs.dev/)** uses **chokidar v5** for its dev server precisely because it is **fast** and battle-tested on large trees. Aligning Testem with that stack reduces surprise for contributors and makes behavior easier to reason about next to other tools in the same repo.
- **Policy:** Keep Testem’s glob include/ignore rules explicit (see **Glob and minimatch**) so config intent stays clear even though the **underlying** watch events can differ from fireworm.
- **Support matrix:** Align Node support with maintained LTS lines and minimum versions required by current dependencies.

### Runtime

- **`lib/file_watcher.js`** — Public `FileWatcher` facade; composes implementation details.
- **`lib/file_watcher_impl.js`** — Chokidar-backed watcher; awaits chokidar’s `ready` before use.
- **Policy / matching** — Centralized glob policy (`file_watch_glob_policy.js`), path/glob helpers (`path_pattern_match.js`, `expand_glob_pattern.js`), and **`is_emfile_error.js`** for EMFILE handling surfaced in the dev UI. See **Glob and minimatch** below.
- **`lib/app.js`** — Wired to the new watcher API and lifecycle.
- **`lib/config.js`** — Adjustments as needed for watcher integration.

### Tests

- Unit and integration coverage for the watcher, glob policy, path matching, app-level file watching, and promise utilities.

### Tooling

- **`eslint.config.js`** — Small updates so ESLint and Prettier stay aligned.

### Documentation

- **`README.md`**
  - Opening paragraphs: what Testem is (**JavaScript test runner**), running tests in **real desktop browsers** vs only Node/headless, **framework-agnostic** support for **any kind of tests** (unit, integration, custom).
  - **Features** list: clarified framework-agnostic bullet and **all major real browsers** bullet; no removal of existing items.
  - **IE and PhantomJS:** Still supported, but **informationally marked deprecated** in the README—continuing to support them through transpilation and polyfills is expected to get harder over time; points readers to `docs/config_file.md` for browser categories.
  - New **File watching** subsection under Development Mode: `src_files` / `watch_files` / `src_files_ignore` / `disable_watching`, cwd + glob policy, **`CHOKIDAR_USE_POLLING`** / **`CHOKIDAR_INTERVAL`** troubleshooting, link to chokidar.
  - Development walkthrough: notes that the dev URL is opened in a **real browser** (Chrome, Firefox, Safari, etc.).
  - **Preprocessors** section: points **`src_files`** at the File watching section instead of duplicating watch behavior.
- **`docs/config_file.md`** — New **File watching behavior** subsection under common options: chokidar watches `cwd`, how **`src_files` / `watch_files` / `src_files_ignore`** form the policy, `disable_watching`, polling env vars; **`serve_files` / `test_page` / `launchers`** remain top-level list items after that subsection.
- **`docs/use_with_jenkins.md`** — Whitelist example uses **`IE11`** (launcher name) with **IE 11** in prose so user-facing docs avoid naming Internet Explorer versions below 11 while keeping a concrete CI example.
- **`docs/browser_args.md`** — Browser list uses generic **IE** (no version in that line).

### Glob and minimatch

Chokidar **only watches** the tree (typically the project `cwd`). It does not include globbing or ignore semantics. Therefore testem cannot delegate full `src_files` / ignore semantics to the watch engine. Instead:

1. **`buildWatchGlobPolicy` (`file_watch_glob_policy.js`)** builds two lists from config: **include** patterns (config file path, default `*.js` in cwd mode, then `watch_files` if set, then `src_files` / default `*.js`) and **ignore** patterns (`src_files_ignore`). That is the single definition of “what counts” for dev-mode watching.

2. **Runtime path filtering (`path_pattern_match.js`)** uses **`minimatch`** to test each changed path reported by the watcher: paths are normalized to POSIX-style (`convertToPosix`) so matching behaves consistently on Windows vs Unix. A path triggers a run if it matches **any** include and **no** ignore (`pathMatchesWatchTarget`). The module is a thin wrapper so the matcher can be swapped later (e.g. **picomatch**) without spreading `minimatch` calls across the codebase.

3. **Filesystem expansion (`expand_glob_pattern.js`)** uses the **`glob`** package when Testem needs to **enumerate** files on disk for the same policy (e.g. `expandWatchableFilePaths`): one resolved pattern at a time, with `ignore` passed through to `glob`, results sorted—aligned with how `Config#getFileSet` already uses `glob` for serving / listing, but kept separate so watcher code does not depend on `Config` internals.

**Summary:** **`glob`** = expand patterns to file lists; **`minimatch`** = match a single event path against include/ignore lists. Together they keep watcher behavior aligned with existing Testem glob semantics instead of relying on chokidar’s built-in glob/ignore alone.

**Why not replace `glob` / `minimatch` in this PR?** We kept them on purpose to **keep the breaking surface smaller**: this change is already large (fireworm → chokidar, app wiring, tests). Swapping glob or matcher implementations at the same time would add another axis of behavioral drift for users. **`glob`** and **`minimatch`** are also **already in the project’s dependency tree**—for example **`Config#getFileSet`** and related code already rely on **`glob`**, so the watcher path reuses the same stack instead of introducing a second glob backend in one release. **Alternatives** (e.g. fast-glob, tinyglobby, picomatch) are explicitly **out of scope here** and listed under **Future work** below.

## Future work

- **Performance / load testing for file watching** — Add benchmarks or load-style tests (large trees, many patterns, rapid churn) so regressions in watcher cost are visible over time. Not in scope for this PR.
- **Swappable glob / matchers** — As discussed above, we did **not** change **`glob`** / **`minimatch`** in this PR; with **`expand_glob_pattern`** and **`path_pattern_match`** as narrow boundaries, follow-up work can try **faster implementations** behind the same policy API and compare behavior and performance in isolation. For **`glob`** expansion, **[fast-glob](https://github.com/mrmlnc/fast-glob)** and **[tinyglobby](https://github.com/SuperchupuDev/tinyglobby)** are plausible experiments; for path matching, **picomatch** (or other matchers) can replace **`minimatch`** via **`path_pattern_match`**.

## User impact

- **Config surface:** Option names and intended meaning for `src_files` / `watch_files` / `src_files_ignore` / `disable_watching` are unchanged; glob policy is centralized as described under **Glob and minimatch**.
- **Runtime:** Treat this as a **major**-style upgrade: verify **Node** against `engines`, and re-check **file watching** in dev mode on your machines (especially Docker, NFS, VMs, or non-default layouts).

## Proposed versioning (for maintainers)

This migration warrants a **semver major**. A practical sequence:

1. **Ship pending 3.x work first** — Cut a **Testem 3.x** release from `master` that includes any other changes already merged there, **without** this PR, so users get a last stable line before the watcher swap.
2. **Publish a beta** — Merge this work and release it as **`4.0.0-beta`** (or similar pre-release tag on npm) so early adopters and CI can pin it explicitly.
3. **Validate in the wild** — **Smoke-test** `4.0.0-beta` on **real projects** that use Testem (dev and CI, different OS/filesystem setups) and collect feedback before declaring the stable API.
4. **Release 4.0.0** — When confidence is high, publish **`4.0.0`** final; document any last-minute notes in release notes / changelog.

This keeps the major bump predictable, gives the ecosystem a clear pre-release to try, and avoids rushing a stable `4.0.0` before it has been exercised on representative workloads.

This PR description, and many changes to documentation in this PR, were generated with Cursor using model Composer 2.0


mansona:
This should have been a breaking change and has had downstream effects even on cases when you're running with Node >20.19

You can see it causing an issue here: https://github.com/mainmatter/ember-vite-codemod/actions/runs/24836882075/job/72699550643?pr=156#step:7:53 and that was run on Node 20.20.2

@johanneswuerbach can we get this reverted and released in a patch. If we want to re-introduce this change it should be marked as a breaking major change

johanneswuerbach:
We certainly test against Node 20 and haven't seen this issue in our CI https://github.com/testem/testem/actions/runs/24666271488/job/72124634384

The call stacks seem to point more towards execa, which would mean this change https://github.com/testem/testem/pull/1947

Looking at your environment, https://github.com/mainmatter/ember-vite-codemod/actions/runs/24836882075/job/72699550643?pr=156#step:7:124 it also seems to run an old v20 minor.
image

mansona:
I mean if it previously worked on Node 18 and this change makes it not work any more (see other PRs) it should still be marked as a breaking change

As for this CI I will do some investigation but it looks by this step that it's using 20.22.2: https://github.com/mainmatter/ember-vite-codemod/actions/runs/24836882075/job/72699550643?pr=156#step:4:11

mansona:
Turns out I had a `.npmrc` with the `use-node-version` setting that was picking an older node version for pnpm commands.

I will comment on the other issue because it looks like other people are having the same problem I'm describing here

PR #1993, opened 2026-04-16
add vite example

Gaurav0:
See included examples/vite/README.md


Gaurav0:
@johanneswuerbach I'd like to publish https://github.com/testem/testem/blob/0b293ea1b721657aa0f20b685e1395fa610d6c32/examples/vite/vite-plugin-testem/index.js as a separate repository and npm package.

Ideally, this should be done within the testem org as vite-plugin-testem I think. I could only do this for you if made a member of the organization and given appropriate permissions in Github and npm. Or you can do that yourself.

Or I could just proceed and publish it separately myself, without any relation to the testem organization. It would still have the same name, vite-plugin-testem, so it's not that big a deal, it's just that I'd be the sole maintainer. Let me know what you decide.

PR #2044, opened 2026-09-03
Prefer node_modules for built-in runners; deprecate Jasmine 1

Gaurav0:
## Summary
- Built-in `mocha`, `mocha+chai`, `qunit`, and `jasmine2` runners now load framework files from `/node_modules` in the project `cwd` when those packages are installed, and keep today’s CDN pins (Mocha 2.3.4, Chai 3.4.1, QUnit 1.20.0, Jasmine 2.4.1) as fallback.
- `framework: "jasmine"` remains Jasmine 1.3.1 from CDN (not an alias for Jasmine 2). Testem logs a deprecation warning on the server and in the Jasmine 1 adapter; Jasmine 1 will be removed in the next major version.
- Examples that use those built-in runners opt in with `mocha`/`chai`/`jasmine-core`/`qunit` so integration runs pick up local files. This is a non-breaking 3.x change (no version bump).
- A followup PR will remove jasmine 1 and all cdn imports. It will also alias framework: "jasmine" to modern jasmine. That is a considerable breaking change.

## Test plan
- [ ] `npm test` (includes resolver unit tests and server HTML assertions for local vs CDN paths)
- [ ] With `mocha`/`jasmine-core`/`qunit` installed in cwd, open a built-in runner page and confirm `/node_modules/...` scripts (jasmine-core 5 uses `boot0.js`/`boot1.js`)
- [ ] Without those packages (or `cwd` that has none), confirm the previous CDN URLs still load
- [ ] `framework: "jasmine"` (default) still serves Jasmine 1.3.1 CDN and prints the deprecation warning
- [ ] `mocha+chai` with local Chai 6: specs stay classic scripts (`var expect = chai.expect`); Chai is loaded as ESM then `loadScript()`
- [ ] Examples `mocha_chai_simple`, `jasmine2`, `qunit_simple`, `hybrid_simple`, `custom_reporter`, `browserify` after `npm install` in each directory
- [ ] Jasmine 1 examples (`jasmine_simple`, etc.) still run via CDN

Made with [Cursor](https://cursor.com)


johanneswuerbach:
Nice, could you rebase?

PR #2056, opened 2026-09-04
Unskip webpack example on Windows

Gaurav0:
- Fix webpack example to work on Windows and unskip it.


johanneswuerbach:
This has now a conflict.

